PR #97, opened 2015-05-19
[Fix #89] [Fix #3] Fix Certain Panel Problem

### LINKED ISSUE #3: Switching fleet around
When I swap girls in the same fleet, the girl I moved gets duplicated in the fleet info until I go to the homepage again. Ex. I have a fleet with Haruna and Kongou, I place Kongou on the first slot, now it shows a fleet with Kongou and Kongou, and it gets refreshed into Kongou and Haruna once I go back to the home

### LINKED ISSUE #89: New Developed Item not registered
i just equipped Type 3 Shell (and it JUST developed) and this happens ?
i wonder why... (i hope you could update the item list after successful scrap/development)
image

ReiFan49:
* Allows quick ship checking for ship swapping purpose.
* Reduces conditional check and transfer new item logging from Show Craft Box to Reactor/CreateItem

### DIFF
--- a/src/pages/panel/classes/Dashboard.js
+++ b/src/pages/panel/classes/Dashboard.js
@@ -169,14 +169,6 @@ KC3.prototype.Dashboard  = {
 			$("#craftModal .equipText").html("You already have <strong>"+countExisting+"</strong> of this equipment, +1!");
 		}
 		
-		// Add the new equipment to local data
-		app.Gears.set([{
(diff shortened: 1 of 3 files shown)


ReiFan49:
should i ask for pull again? forgot that "craftData" is not "response.api_data"

dragonjet:
ah I already fixed it on a commit afterwards

Issue #110, opened 2015-05-25
Admiral Stats Screen Translation Overlay


URSvAir14:
Whenever I look at the admiral stats sreen I honestly have no idea what most of it means, so whenever I want to go onto it I need to cross reference with the translation image at http://img3.wikia.nocookie.net/__cb20150412115518/kancolle/images/d/d6/Status_screen_translated_12-04-2015.png . I was wondering if it was possible to have an overlay pop up, similar to how it does with the quest screen, with these translations whenever you visit the admiral stats page.

dragonjet:
closing issue. feature was added by @tmerta

dragonjet:
Will need to re-import this due to v010 overhaul

PR #123, opened 2015-06-06
[#Feature 110] Admiral Stats Screen Translation Overlay

tmerta:


tmerta:
Thanks! Since I screwed up the issue link it doesn't seem like the related issue, #110, was updated correctly. Sorry about that.

PR #143, opened 2015-06-18
[Fix] Compass result for enemy fleet doesn't work

bamboo3250:
I noticed after moving the Abyssal images to photobucket, the compass cannot show the enemy fleet image anymore so I made a fix for it. Now it looks like this.

untitled

### DIFF
--- a/src/library/classes/Assets.js
+++ b/src/library/classes/Assets.js
@@ -1,5 +1,6 @@
 KC3.prototype.Assets = {
 	cdn: "http://i708.photobucket.com/albums/ww87/dragonjet25/KC3%20Ship%20Icons/",
+	abyss: "http://i708.photobucket.com/albums/ww87/dragonjet25/KC3%20Abyss%20Icons/",
 	_ships :[],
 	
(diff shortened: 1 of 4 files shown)


dragonjet:
looks great :thumbsup:

bamboo3250:
I think so, that __background:rgba(255,255,255,.7);__ was from the enemyFleet previously then I moved it to __transparent-container__.

Crazyteitoku:
@bamboo3250 :+1: on the fix - I was sort of going to remind dragonjet via a thread about a new kc3 kai layout on [wikia](http://kancolle.wikia.com/wiki/Thread:273592#4)

Looks like we can wave bye to the old layout and say hello to the new:

bamboo3250:
I think there is no file named 'library/managers/TimerManager.js' at the moment. Maybe @dragonjet will add this later? Have any idea?

dragonjet:
whoops! fixed it on new commit

Crazyteitoku:
Thanks! latest commit did address the issue.

Issue #58, opened 2015-05-09
Battle conditions reminder


AlanSchaeffer:
Expanding on the battle prediction proposed before, I'd like to suggest something that might be useful too to those that do not wish to see the prediction itself.
There could be a panel that would be shown during battles to view those parameters that are only shown briefly at the start of the battle, in case I miss it because I wasn't looking at the game, for example. It could show the engagement type (the green/red T thing etc), which side got air superiority, and that scout thing that happens which is said to do stuff, but I have no clue what it does (even though it is on the wiki lol).

bamboo3250:
I have a liking to this suggestion :+1: Probably this is what I will do next.

bamboo3250:
@dragonjet please close this issue. I forgot to put "fix" in the comment.

bamboo3250:
@AlanSchaeffer I added the feature u requested here #148

dragonjet:
will need to reimport for v010

bamboo3250:
okay, no problem :+1:

dragonjet:
by the way, we will have a special interface for battle mode, when user starts a sortie. see new issues I made under horizontal battle mode.

The compass and enemy list will not be popup boxes but will have their own always-present box in the dashboard. though I still haven't started on this "battle mode". I'll notify you again on this thread when the box is ready to be filled.

dragonjet:
Forgot to close but has been addressed on
9d3e2b27b5902533c5f316e326ca70515e74cd69
f7ec1dc38d95d5503550eb25a7a4eaf6f52c9af7
5e50f5c0edbb3e236e233fece35c11553b281f06

PR #148, opened 2015-06-20
[Feature #58] Battle conditions reminder

bamboo3250:
Added a modal to show some statistic of the battle when engaging the battle. #58 

untitled


dragonjet:
wow this looks very good :+1:

dragonjet:
question, the red plane is the plane count right?

bamboo3250:
haha yes, those are plane counts, not fighter power. The first column is the number of planes that involve in Fighter Combat (only planes that have AA stat). The second one is the number of Dive and/or Torpedo Bombers are shotted down and remaining.

Issue #157, opened 2015-06-27
[v10] Dashboard: Vertical (Normal Mode)


dragonjet:
a new interface for vertical orientation, with 800px width

dragonjet:
Addressed on latest pull requests by @bamboo3250

PR #179, opened 2015-07-04
[Feature #157] [v10] Dashboard: Vertical (Normal Mode)

bamboo3250:
Solved #157

@dragonjet Here is the Normal Mode for Vertical Layout. Just tell me if you want any amendment. :smile: 

untitled

### DIFF
--- a/src/library/managers/TimerManager.js
+++ b/src/library/managers/TimerManager.js
@@ -35,6 +35,22 @@ Container object for timers for expedition, construction, and repair
 		repair :function(num){ return this._repair[num-1]; },
 		build :function(num){ return this._build[num-1]; },
 		
+		updateTimerElement : function(eMap, rMap, bMap){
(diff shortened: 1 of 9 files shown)


dragonjet:
wow that's very fast! and it looks great! just wow, done better than i could have

bamboo3250:
Haha thanks :grin: To prevent conflict, for now I will focus on completing Vertical Layout first. After that I'll take a look at existing issues related to the panel (both orientation).

SmokeyTheButt:
So when I go into battle mode on the vertical layout, I get this odd cutoff. 

audioclip-1434123331

bamboo3250:
I haven't done Battle Mode for Vertical Layout yet so it's expected. I'm going to work on it today. Thanks for reporting.

SmokeyTheButt:
Ah okay.

PR #200, opened 2015-07-10
[Fix #192] Contact on KC3 battle mode not work properly

### LINKED ISSUE #192: Contact on KC3 battle mode not work properly
I re-download the kc3kai from github, but it still show that its have NO contact, 

contact2

hope someone would try to test this

bamboo3250:
Fixed #192

Now it works. I can even show the type of plane that has Contact xD

contact


McDohl:
good job :+1:

dragonjet:
whoa really XD I was near surrender for this but it's just the values are -1 vs everything else, and not 0 vs 1. :laughing:

PR #213, opened 2015-07-11
[Feature] Increase the height of Vertical Layout

### LINKED ISSUE #209: Visual bug on Quest box overflow
Looks like the sorties was simply not counted in Sortie Quest 1, 2 and 3.

Also, the "Daily Equipment Modernization" text spills over to new line in the quest list.
questbug

bamboo3250:
I just simply added more classes (e.g. big and small) into HTML and CSS. For now, it will force user to user big version of the layout but later we can add options to choose big or small. Just replacing `big` by `small` in HTML will work.

btw, could anyone having big screen use this version and send the picture to me? I don't have big screen and it's hard to imagine.

Normal Mode
normalmode

Battle Mode (Single Fleet)
battlemodesingle

Battle Mode (Combined Fleet)
battlemodecombined

ps: this PR also fixed #209

### DIFF
--- a/src/pages/devtools/themes/default/vertical/vertical.css
+++ b/src/pages/devtools/themes/default/vertical/vertical.css
@@ -1760,6 +1760,7 @@
 	height:25px;
 	line-height:25px;
 	float:left;
+	margin:12px 0px;
 }
 #v .big .battle-ship .ship-gear img {
 	width:25px;
--- a/src/pages/devtools/themes/default/vertical/vertical.css
+++ b/src/pages/devtools/themes/default/vertical/vertical.css
@@ -2935,7 +2935,7 @@
 	font-weight:bold;
 }
 
-#v .battle .battle_resource {
+#v .battle.small .battle_resource {
 	width:130px;
 	height:30px;
 	margin:5px 0px 3px 5px;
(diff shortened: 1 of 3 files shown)


bamboo3250:
@dragonjet  This is just my personal request but could you update this pack as soon as possible? I want to read feedbacks from other players to improve the interface further.

McDohl:
my feedback, 450px is just not pretty to see on my end.

it seems too big for me, had to adjust the panel below right after gamescreen like this, else will have scrollbar for dashboard.

newvertical

i tends to use split screen mode since, i prefer firefox than chrome. Chrome only for kancolle. as you can see on below pics. it will have scroll bar on dashboard.

verticalwindowed

my screen resolution : 1920x1080
if splitscreen, chrome auto resize it to 944x480

i think thats all my feedback

McDohl:
my feedback, 450px is just not pretty to see on my end.

it seems too big for me, had to adjust the panel below right after gamescreen like this, else will have scrollbar for dashboard.

newvertical

i tends to use split screen mode since, i prefer firefox than chrome. Chrome only for kancolle. as you can see on below pics. it will have scroll bar on dashboard.

verticalwindowed


my screen resolution : 1920x1080
if splitscreen, chrome auto resize it to 944x480

i think thats all my feedback

bamboo3250:
@McDohl Thanks for your feedback. The screenshot is quite valuable to me xD It seems 420px would be the best fit for 1920x1080. I'll try to find a way to cut off that 30px. That is the best I can do for 1920x1080 users.

bamboo3250:
@McDohl If possible, could you check this one #222 after it gets merged?

McDohl:
sure, would try it asap

bamboo3250:
Thanks :+1:

dragonjet:
merged

PR #222, opened 2015-07-12
[Feature] Size option for Vertical Layout

bamboo3250:
I added 2 buttons on the right side of the Dashboard (only Vertical Layout) to select the size for it.

__Small (300px)__
battlemodesmall

__Big (420px)__
battlemodebig

We also can change it in Settings Page. 
settings

### DIFF
--- a/src/data/translations/en/settings.json
+++ b/src/data/translations/en/settings.json
@@ -259,6 +259,17 @@
 				}
 			},
 			{
+				"id" : "pan_size",
+				"name" : "Panel Size",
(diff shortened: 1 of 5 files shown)


McDohl:
thanks for the hardwork, had no more complaint, no more scroll bar on both windowed split screen, the LoS info also fixed :+1:

bamboo3250:
@McDohl Thanks for feedback xD I'll create an issue for those and investigate them when I have some more ships.

PR #226, opened 2015-07-13
Various fix

McDohl:
fixed for https://github.com/dragonjet/KC3Kai/issues/224 or rather https://github.com/dragonjet/KC3Kai/issues/188 for vertical. the exit battle mode is not fixed since too lazy to look at combined fleet code :sweat_smile: @bamboo3250 

added a color that if ship is being repaired, instead of red, the HP becomes blue, on vertical.js


bamboo3250:
nice fix :+1: :smile:

Issue #229, opened 2015-07-13
[Translation: Japanese] Settings


dragonjet:
**File**
`\src\data\translations\jp\settings.json`

**What to edit**
* Text on `"section": "HERE",`
* Under `"contents" : [`
  * **Do not edit** `"id" : "xxxx",`
  * **Only edit** `"name" : "HERE"`
  * Inside`"options" : {`
    * Only edit the *value* which usually follow the format `"attribute" : "value"`
  * Inside`"choices" : [`
    * Each line usually looks like this, `[ value, "label" ]`, only edit the *label*

**Others**
* Please review the machine-translated items at the beginning

PR #238, opened 2015-07-13
[Translation: Japanese] [Settings] edited (translated) on 警報 (alert) section.

nicky008:
From #229. Alert section done. Trying to do the later paths if I could wake up early today. :v


dragonjet:
:+1:

PR #252, opened 2015-07-14
[Translation: Indonesian] Finished Settings and Record

hufuhufu:
Finished translating settings.json and record.json as an additional


dragonjet:
:+1:

PR #277, opened 2015-07-16
Improve readability for equipment totals

McDohl:
changed the font size to 14px and made it bold type

eqcount

https://github.com/dragonjet/KC3Kai/issues/275

PR #280, opened 2015-07-16
Update gears.css

McDohl:
https://github.com/dragonjet/KC3Kai/issues/275

edit : will try lol


bamboo3250:
I think you can push these commits to #277 to reduce the work for @dragonjet :sweat_smile:

Issue #260, opened 2015-07-15
Quest Reset


JD-Kurotoshigami:
separate the Daily/Weekly quest reset buttons into Quest Types: Daily Sortie, Daily Expedition, etc.

dragonjet:
eeeh what for jede?

JD-Kurotoshigami:
earlier... I pressed the daily quest reset since it still had records of the previous day, but I was doing some dailies that time so even the ones i am doing are reset :'D

LostLogia4:
Wouldn't it be easier to code it to reset the tally once the quest is cleared?

dragonjet:
@bamboo3250 talked about it on another thread i forgot which

bamboo3250:
I think this is a new proposed feature, not the one about automatically resetting quest list.

It's like separating Daily reset button into 5 kinds of buttons for Daily Sortie, PvP, Resupply/Repair, Expedition and Modernization.

dragonjet:
ah yes, it was what JD was suggesting, but I think what LostLogia4 said can suffice

bamboo3250:
oh right. I misunderstood what LostLogia4 said xD I'll finish that soon.

LostLogia4:
One thing to note though, the tally still have to be reset at certain intervals, like every 5AM JST everyday for daily quests.

bamboo3250:
Unless we have a mechanism to detect the time and time zone, it's impossible to reset the quest at 5AM JST since the API response doesn't provide enough information. Hence we still have to rely on some workaround, which seems to have the same effect but still potentially has bugs inside.

dragonjet:
I tried to do this before, using server time included in the response header. detecting server time is not a problem, it's a problem of "shall we reset it now? have we passed 5am since the last time he did something?". my old implementation included a timestamp which we store locally every time an API is called. On the next API call it will check if `oldServerTime < 5am < thisServerTime`, which is painful to do by the way - but I postponed the implementation because it really became harder to detect weeklies.

if anyone is up to the task I can provide my old attempt and can continue where I left it

bamboo3250:
@dragonjet omg, I didn't notice that server time in GMT :scream: I would like to see that :sparkles:

dragonjet:
```
// If there was a last quest time
if(typeof localStorage.lastQuest != "undefined"){
	// Convert into Date objects
	var dateQuest = new Date( localStorage.lastQuest * 1000 );
[...]

PR #282, opened 2015-07-16
[Feature #244] Automatically reset counter when navigating quest list and clearing the quest

### LINKED ISSUE #244: [Feature] Automatically reset counter when navigating quest list and clearing the quest
@dragonjet I would like to work on this feature. Is that okay?

bamboo3250:
resolved #244 

Due to the finding in #260, now it will reset daily quests when 5AM JST of a day has passed at the first time. For weekly quests, it is on Monday 5AM JST, and for monthly quest, it is on the first day of the month at 5AM JST.

In addition, this implementation also includes auto-adjusting-counter functionality. For example, because of some reasons, the counter counted wrongly, showing 1/5 while the game screen show 50% of the quest has been done. In this case, it'll increase the counter to 3/5 to satisfy the progress status of the quest. That means it will look more correct but not always accurate. It is for reducing the awkwardness of this kind of situation.

For testers, after pulling the code, please restart the panel and open console for the console like this. 

quest-reset

`Reset Daily Quests` is the time to reset daily quest.
`Reset Weekly Quests` is the time to reset weekly quest.
`Reset Monthly Quests` is the time to reset monthly quest.

Please check whether those lines printed above are correct or not. (checking at your local time is prefered)


McDohl:
i dont know how to show that reset xxxx quest.

so far from what i tried, auto adjust tracking counter is working :ok: 
dummo

bamboo3250:
@McDohl You try selecting KanColle Tab then use `Ctrl + Shift + I` on that tab. It will show another panel for the KanColle Tab only.

dragonjet:
its a devtool for a devtool :v

McDohl:
devtools

NaN invalid date o.O

devtoolsception :v

Crazyteitoku:
Yeah reporting the same thing on my end too.

Looks like there needs to be another way to obtain date and time.

dragonjet:
eh, @McDohl you're using dark theme? :open_mouth:

Crazyteitoku:
@dragonjet he too is also using the dark theme :stuck_out_tongue_closed_eyes:

McDohl:
yeah, but seems like its not working :stuck_out_tongue_winking_eye:

Crazyteitoku:
It is a feature request really, pretty sure it is low priorities right now (or otherwise put on the backburner) due to the lack of interests. Then again, you can see the demand for the dark theme is there - and that it is increasing.

McDohl:
i saw at settings theres dark theme, n decided to use it but nth happen (hardly open settings anyway :laughing: )

that aside, its really minor thing so i let it slide. 
I think KC3改 should put strategy room features top priority atm cuz event is around the corner.

bamboo3250:
I solved the problem. I will create a pull request for that right now.

Crazyteitoku:
I have loads of kc3 kai tabs open on my end, it is always: KC3 kai, strategy room, settings and about tabs :open_mouth: 

@bamboo3250 wow! that's quick. I was poking around randomly with my noob coding skills and saw that it uses similar functions when crafting or lsc. Strangely enough that on this Questmanager.js part it requires more work. :+1:

bamboo3250:
Hopefully @dragonjet will merge the new one #284 soon. This kind of bug is really not good for heart :sweat_smile:

Crazyteitoku:
@bamboo3250 @dragonjet  is around, pretty sure he'll be quick to merge that in. Excellent work! at least quest tracking feature is getting closer to completion. :+1:

McDohl:
console
@bamboo3250  there

dragonjet:
merged

Crazyteitoku:
I just did a git pull on this and still getting the same issues on my end. I think it is becoming platform specific, might have a poke at the codes and see where I can point out where it maybe is wrong.

git revision: 8bd8464

dragonjet:
pull again, that head still doesnt have @bamboo3250's fix

Crazyteitoku:
Ah thanks! @dragonjet

bamboo3250:
@Crazyteitoku yeah, the quest is almost completed now, just the quests about sinking a certain type of ship left. That would be a hard one :sweat_smile: 
@McDohl thanks, it looks correct :+1: 
@dragonjet thanks :laughing:

dragonjet:
by the way, since maintenance later 11am JST, I'll be early uploading v11 on Webstore by around 1130. We can't test things if we cat bomb so :laughing:

Crazyteitoku:
@bamboo3250 I sort of wondered about that, I do recall @dragonjet said that the sinking of certain types of ships (namely CV/L or transports) is a no fix because the server doesn't send those data to the client for some reason :open_mouth: 

If server is able to send results, i.e. A or S and assuming if S result would it able to automatically count from that? I mean A is much harder to count and may require manual intervention but otherwise S is more possible considering that the player would have destroyed all the said enemy ships on the map? This is also a bit obvious but yeah.

dragonjet:
that's true CT and is a possible temporary fix, but during orel which 3 subs rarely do an S, it will be super inaccurate. the real solution is to do the battle prediction where we calculate the attacks and damages ourselves

bamboo3250:
@Crazyteitoku dragonjet is right. I also aim to calculate the damage that a ship received, together with its HP and the type of ship. That would be possible to solve the problem but will required much investigation and time to do so.

Crazyteitoku:
@dragonjet I normally bounce between orel and 2-2 for that, at least for Bw4 quest: ろ号作戦 (ro gou sakusen) whereby one needs to sink 50 transports and that is after the Bw1 quest: あ号作戦 (a gou sakusen).

@bamboo3250 yeah seems like that would be a pain and interestingly enough that the server doesn't offer the information upfront.

dragonjet:
We can use EO's battle emulation as basis: [Battle phases where damages are calculated](https://github.com/andanteyk/ElectronicObserver/tree/master/ElectronicObserver/Data/Battle/Phase)

You can browse their library for more info

Crazyteitoku:
@dragonjet thanks :+1: but I'd probably be clueless looking at their codes. The last time I recall the backend stuff on EO relies on fiddler core which is the same as KCV and well different programming language in short.

bamboo3250:
@dragonjet can we have a BattleManager.js like theirs :sparkles:

dragonjet:
hm usually, I use "managers" if that class contains a list of its objects `ShipManager.js` for `Ship.js`, etc. For sorties and battles, that's our `SortieManager.js`. My original plan is just to add it to `\objects\Node.js` then just add `KC3Node.prototype.simulate` - but looking at it, this will involve many many functions, separating each phase, and calculators. hmm.. what about another object `Battle.js`, then inside Node.js we `myBattle = new KC3Battle()` then `myBattle.simulate()`?

PR #290, opened 2015-07-17
Localization enhancement

Diablohu:
All translations will now extend based on English version, thus you can
remove all "non-text" content from translation json files. And even
specific text dosen't exist in given language json file, it will be
showed in English on page.

NEW FUNCTION: $.getTranslationJSON( language, filename, callback )
* in /src/library/modules/Meta.js

All exist source code related to translation are all converted to the new one.

You can check /src/data/translations/scn/servers.json for example.


dragonjet:
this is good enhancement, excited to test it out~

nicky008:
:+1: This...

bamboo3250:
This helps a lot :+1:

PR #292, opened 2015-07-17
[Translation: Indonesian] Finished Servers, Terms, and Developers

hufuhufu:
Added servers.json, developers.json, and terms.json, also added 2 languages on settings.json, and deleted record.json
Update: terms.json 17 July update

Also how can I delete record.json on your branch?
Will it be deleted if I delete the one on my branch?
Is that how it's works?

Sorry so many question xD


nicky008:
Delete that record.json on your branch first then commit your work. :v:

hufuhufu:
Ok great thanks :smiley:

PR #307, opened 2015-07-18
ExpeditionHelper quick fix

Javran:
A quick fix to make it work with new API, keep in mind that this messes up the current layout a little bit.

### DIFF
--- a/src/library/helpers/ExpeditionHelper.js
+++ b/src/library/helpers/ExpeditionHelper.js
@@ -1,351 +1,373 @@
 (function(){
-	"use strict";
-	
-	window.KC3ExpeditionHelper = {
-		preconds: {
-			1: function(info) {
-				return info.shipCount >= 2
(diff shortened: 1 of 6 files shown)


bamboo3250:
Could you take a screenshot while we wait for dragonjet to merge it?

Javran:
Here you go, this is the horizonal one:

2015-07-18_004330_3840x1080_scrot

And vertical one:

2015-07-18_004425_3840x1080_scrot

Basically I just `JSON.stringify` stuff, not really know the best position to place it.

bamboo3250:
That's good. We can temporarily put it inside the white box below. You can leave it to someone else to handle it properly later :+1:

Javran:
I see, thanks xD

dragonjet:
ah I've also put invisible `div`s on that white box so that it will interchange between equipment crafting and expedition helper, that white box was also meant for this xD

dw I'll do the horizotal box for it~

Issue #263, opened 2015-07-15
Shiplist in Japanese. Ship Types too long


dragonjet:
*Japanese Translation : The Kanji in the Ship List messed, the "div" is too narrow. I think it needs to be expanded (or decrease the font size). -Thanks-* - [Fujisuke](http://kancolle.wikia.com/wiki/User_blog_comment:Dragonjet/KC3%E6%94%B9/@comment-Fujisuke-20150715101134?permalink=283262#comm-283262)

nicky008:
Some corrupted pinyin words (I think because of the masking problem) are overlapped the normal one.

dragonjet:
ah it should've been the font size only, the overlapping words are actually the words above them being multi-line

kyonmiriam:
Not only Japanese Translation but also the Chinese Translation(both Simplified Chinese& Traditional　Chinese).I think this problem will appear in every hieroglyphs languages...including Korean. So why not let a single line contains more characters instead of decrease the font-size? Ah...maybe shorten the  translated texts can also solve the problem. After all, two Kanji can represent too many things...:flushed:

dragonjet:
true, our interface issues are growing, especially with localized languages and unicode characters with varying length. coincidentally, I just started a topic about re-thinking our UI options on the other thread #312, though that one is mainly for the F12 panel. This one is for the strategy room shiplist, and the columns there, admittedly are getting narrower and cramped. With the recent addition of the morale column, and without main layout not being responsive, we're moving to a dead end.

I hope we do have shorthands for ship types in other languages though, like for Japanese, instead of the 5 kanji chars, we should use the 1 or 2 shorthands found on the in-game ship thumbnails

nicky008:
Maximum char (kanji, pinyin, hangul etc.) for here should be 3. :ok:

dragonjet:
ah if it's 3 then we can support that as well since english also is max 3

c933103:
i think this should be closed?

dragonjet:
ah yes thanks for noticing

PR #316, opened 2015-07-18
[Translation: Japanese] [Stype] Reducing number of chars for #263.

nicky008:
From #263, each ship type word (kanji) in Japanese will has only 3 chars in maximum number now.


src/data/translations/jp/stype.json
```diff
@@ -2,23 +2,23 @@
...
-    "軽巡洋艦",
-    "重雷装巡洋艦",
-    "重巡洋艦",
-    "航空巡洋艦",
+    "軽巡洋",
+    "重雷装",
+    "重巡洋",
+    "航空巡",
     "軽空母",
     "F戦艦",
```
bamboo3250: Do the japanese use any kanji for "Fast" in Fast Battleship or just F?

Crazyteitoku:
@bamboo3250 I know in the game technically it is meant to be 高速戦艦 (kou soku sen kan) which literally means fast speed battleship when it comes to describing fast battleships. Not sure elsewhere however.

nicky008:
Frankly, their is this word "高速戦艦", directly mean "fast battleship". But in Kancolle, the devs still retained to use "戦艦" for any fast battleships here for now.

I personally want to use "高戦艦" for fast battleship in my "3 chars rule" if most of everyone agreed.

bamboo3250:
If that is the case then we should stick with F戦艦. I cannot find any document that uses the word 高戦艦 or 速戦艦 for fast battleships. Perhaps we cannot separate the word 高速.

LostLogia4:
@bamboo3250 is right. "高" means tall or high and has nothing to do with speed. "速" means speed, but it has no sense of magnitude on its own, so individually neither kanji could represent "high speed" by itself.

dragonjet:
I guess I'm the only one with zero japanese :3

nicky008:
@LostLogia4 @bamboo3250 Thx for your suggestions...

bamboo3250:
No need for that. You did it correctly from the start though :stuck_out_tongue_closed_eyes:

c933103:
wouldn't 
 "軽巡",
   "重雷巡",
  "重巡",
    "航巡",
work better?

nicky008:
@c933103 Nice suggestion, I had (at least once) thought these words should be in this (巡) from as suffix.

PR #334, opened 2015-07-19
Chinese translation v.1.2 kololz

kololz:
Added a readme to all translators.
Also finished 96% quests.json.
Just around 30 sortie quests left.


dragonjet:
hi @kololz , it seems there are usually unnecessary double newlines in many quest descriptions. is this intended or something along the way? unfortunately these newlines break JSON standards that make the build fail

https://travis-ci.org/dragonjet/KC3Kai/builds/71655948

PR #336, opened 2015-07-19
[Feature #254] Blur HQ name feature

### LINKED ISSUE #254: Blur HQ name feature
ability to click on HQ name, and it will become blurred for those who are cautious in doing screenshots. double tap will possibly also show up an overlay to cover the HQ name on top-left

bamboo3250:
resolved #254
Sorry for cannot not make it look like holy light :sob: 

censor


nicky008:
This is far enough than I though... :godmode:

PR #363, opened 2015-07-23
Fix ships name

McDohl:
https://github.com/dragonjet/KC3Kai/issues/355


bamboo3250:
Oh nice, thanks :blush:

McDohl:
oh i think other translator should add add / translate the new radars :stuck_out_tongue_closed_eyes:

bamboo3250:
lol I didn't know that. Probably we should create an issue whenever there is new update and tag all translators there :worried:

PR #409, opened 2015-07-26
KC3Translation.getJSON - recursive extend

Diablohu:
the extend method for the new KC3Translation.getJSON() is not perfect since it use non-recursive extending. this will fix that

now non-text content can be safely removed from all translation files except `terms.json`


nicky008:
I'm waiting this. :+1:

PR #427, opened 2015-07-29
KanColleHelpers refactoring and a bug fix

Javran:
* No API change except that only `currentHp` fields are available. Let me know if you are still interested in damage took in each stage of the battle, they can still be extracted from battle data,
but for now getting access to this functionality is not js-friendly.

* Just find a crash while doing daily exercise, turns out [sometimes the array member of `api_df_list` could be `-1`](https://github.com/Javran/purescript-kancolle-helpers/blob/master/test/BattleData.js#L194), it seems simply ignoring them won't affect the result.


dragonjet:
`currentHp` is the HP remaining after damage calculation?

Javran:
@dragonjet yes, you are right.

Issue #437, opened 2015-07-30
[Bug] Fleet tab in strategy room goes kaput


LostLogia4:
If you had scrap/modfodder any of the ships currently in the 1st fleet and go to the current fleet tab, the strategy room would hang and give this error:

Uncaught TypeError: Cannot read property 'substr' of undefined

ReiFan49:
i guess this is caused because there's no update while destroying the ship (with any way) to the fleet data.. using fleet swapping technique might works hmm (gonna check on it)

Issue #429, opened 2015-07-29
HQ Exp Information clashing


ReiFan49:
well, about what i've asked (and i've done it too) before (at #374) there's some information clashing that given from the script.
```javascript
/** on SortieManager.js **/
resultScreen :function( resultData ){
  if(this.currentNode().type != "battle"){ console.error("Wrong node handling"); return false; }
  this.hqExpGained += resultData.api_get_exp; // (<)
  this.currentNode().results( resultData );
  PlayerManager.hq.updateLevel( resultData.api_member_lv, resultData.api_member_exp); // (>)
}
```
```javascript
/** on orientation.js **/
BattleResult: function(container, data, local){
  if(KC3SortieManager.currentNode().type != "battle"){ console.error("Wrong node handling"); return false; }
  $(".battle .battle_current", container).text("RESULTS");
  var thisNode = KC3SortieManager.currentNode();
  
  // If EXP left exceeded by gained EXP on sortie
  while(KC3SortieManager.hqExpGained >= PlayerManager.hq.exp[1]) { // (<)
    KC3SortieManager.hqExpGained -= PlayerManager.hq.exp[1];
    PlayerManager.hq.exp = [0,KC3Meta.exp(++PlayerManager.hq.level)[0],0];
    this.HQ(container, {}, local);
  }
  $(".battle_hqexpgain", container).css({width: Math.floor((function(){ // (<)
    return (PlayerManager.hq.exp[2] + Math.min(PlayerManager.hq.exp[1],KC3SortieManager.hqExpGained)) / KC3Meta.exp(PlayerManager.hq.level)[0];
	})()*60)+"px"});
  // rest of code is omitted
```

now i confused, because the piece of code that marked with `(<)` gives delta of exp gained during sortie, while `(>)` gives the actual HQ status.

sounds like, we need a toggle? 
that separates the work of `(<)` and `(>)` (also, `KC3Network.trigger('HQ');` included as `(>)` feature, because on `(<)` only done whenever HQ level up)

**PS: my commit is only to check is it good enough or not to be a toggle**

bamboo3250:
I understand this conflict because I wrote the (>) part. My point is that they both have the same result. Just the ways they calculate are different. If we only need to show the HQ level and exp, mine will be suffice. But later if we want to add animation to show the increasing exp, we have to use your work. So it depends on the goal of KC3Kai to decide which one should be used. Having some animation in the panel would look really cool, anyway :grin:

[...]

PR #438, opened 2015-07-30
[#429] [#437] HQ Info Delta Depre'd + Remove Zombie Ships on Fleet that Erroneus

ReiFan49:
zombie ships? yes. error infested data.
(actual error infested)


Crazyteitoku:
```
build/tmp/pages/settings/settings.js
if ((response[sctr].contents[cctr].hide || 0) == 0) // hide "private/deprecated" settings
                                                                   ^ Use '===' to compare with '0'.
```

PR #443, opened 2015-08-01
Add Screenshot button on Battle Mode for vertical

McDohl:
300px  :
single 300

combined 300

420px:
single 420


dragonjet:
oops wrong button :laughing:

PR #448, opened 2015-08-01
[Feature #29] Battle Prediction

### LINKED ISSUE #29: Battle Prediction
Show results, damaged and ending HPs right when battle starts.

This will be applied to the mini-HP bar on the panel

bamboo3250:
resolved #29 

It will show HP before and after battle with abyssal ships sunk. 

screen shot 2015-08-02 at 7 43 01 am

It can be switched off by unchecking the option Battle Prediction (It was Battle Environment before).

screen shot 2015-08-02 at 7 48 59 am

PS: this works in both sortie and PvP
PS2: PvP Night doesn't update because the `PvPNight` is still empty.


dragonjet:
NICE :+1:

Issue #493, opened 2015-08-06
About Battle Prediction, Damage Control, and Sunk!


ReiFan49:
ok, let's start a darker topic.

### scroll down if it's hard to understand (sorry)
maybe i should say this as feature request, because damage control/goddess can prevent any ship from sinking on that battle.
as we know, damage control cancels the damage and heals the kanmusu about 20% of the HP

**Start Battle with Taiha Condition**
![image](https://cloud.githubusercontent.com/assets/9847780/9109675/99ff8b38-3c61-11e5-93f3-a7aa333fa65e.png)
**Kuso**
![image](https://cloud.githubusercontent.com/assets/9847780/9109681/abc7c22c-3c61-11e5-9d28-8827def65a71.png)
**After Damage Control Consumed**
![image](https://cloud.githubusercontent.com/assets/9847780/9109701/d505eab0-3c61-11e5-949c-26d1033e4909.png)
*these pics are taken during last E-4 Final Kill, no API data recorded and only videos exists to represent it*

nah at the moment, i'm gonna put some stamp on the kanmusu which is sunk or knocked out during PvP (PvP knockout = sunk in real battle)
however, before implementing that, i need to make sure the kanmusu that started battle with **heavy-damaged state did not bring a damage control**, if the kanmusu brought a damage control during the heavily-damaged state, as i said, the HP recovers back to 20% (sad that, from 2 attempts, **all of my experiment kanmusu have a zero remainder of 5 of its maximum HP, so i cannot ensure whether the devs rounded it down or not**), and further calculation should be treated as normal state of kanmusu (which is cannot sunk).

previously, i want to check a knocked out kanmusu on PvP using such technique, however fails ._.
![this](https://fbcdn-sphotos-d-a.akamaihd.net/hphotos-ak-xaf1/v/t1.0-9/11822736_10204647464533150_5011702513831341999_n.jpg?oh=9306ed8cb8f33e685c5a74845de664a3&oe=5653E370&__gda__=1448317380_6cf3d7fbde6133d9b12f122d243a6446)
*just remember, 1HP on screen is different on API data*

### **tl;dr ver**
what i mean in here,

battle prediction/calculation might break if there's a damage control (or similar) equipped to the respecting kanmusu that starts the battle with badly damaged condition (as stated above).

(PS: poor ryuujou, i'm sorry for such things)

bamboo3250:
[...]

Issue #473, opened 2015-08-03
[Suggestion] Gained Experience Relocation


ethrundr:
As for current version, the total experience gained from each sortie or pvp is located under the total required exp to the next level:
![screen shot 2015-08-04 at 06 46 51](https://cloud.githubusercontent.com/assets/13522800/9048521/c899c6fa-3a74-11e5-9d10-e185852c29ad.png)

However, having three numbers (required exp, gained exp, max eq slot) stacking up without any space looks too cramped over there. Therefore, I suggest to relocate the gained exp numbers, to the space above the total required exp, since not a lot of player who will have a name too long that would interfere with the number being there.

ps: About the max ship slot on my screenshot, I was late on loading the Panel, so the max slot was not updated to its actual amount, which supposed to be 240. Please disregard it.

dragonjet:
tbh I haven't followed the HQ exp enhancements earlier, I know @ReiFan49 added an option to switch between which of the exp values will show, so I guess we don't need the -75 below it, since it's the same as switching into "total gained" from settings?

bamboo3250:
I think it's still useful to know how many exp we get in the sortie. Moving above seems better to me.

ReiFan49:
sounds like this?

**16 px from top (tested with 12 full-width characters)**
![image](https://cloud.githubusercontent.com/assets/9847780/9108697/181ba2b6-3c5a-11e5-8324-8ab550f6c9c5.png)

Issue #472, opened 2015-08-03
[Suggestion] Improvement on Rank in Panel and Strategy Room


ethrundr:
In the Admiral Panel:
The rank should be clickable, to switch between: 
- Rank Title (Admiral, Marshal Admiral, etch)
- Rank Number (1000, 500, etc)
- Rank Points (as mentioned by @dragonjet in #122 )

In the Strategy Room:
The "Rank" row will be divided into three columns, each will consist the informations mentioned before: Rank Title, Rank Number and Rank Points.

dragonjet:
I'm afraid the second one is not possible, and the user needs to go to the actual ranking page in-game to know his rank, which is not available all of the time. That said, it also will not update automatically if you go up or down the ranks and will need to revisit the in-game rank pages again. This means putting it on the panel will not be effective since the user will be seeing it in-game anyways. This is also ineffective for strategy room (which use more limited data, only based on what panel saves on `localStorage`) thus will become outdated 99% of the time.

The third one is possible though, when we do apply Flonnaru's suggestion in the strategy room.

nicky008:
Addendum: Not only be clickable but also have a hover box shown more details about how to cycle (look) more data too.

dragonjet:
hm true, people usually dont explore the panel enough with the mouse to even know which are clickables and which have hovers. I even think 50% do not know the elos icon is clickable to switch between formulas

nicky008:
And they also have this problem in the the HQ exp enhancements earlier too. :hushed:

ethrundr:
actually I know that the HQ exp is clickable just after read your reply, @nicky008  :laughing:

dragonjet:
wait the HQ exp is clickable?

bamboo3250:
@dragonjet  I think you merged #458 :laughing:

ethrundr:
@dragonjet yep, clickable: EXP to next Lv, EXP gained this Lv, Total EXP
![screen shot 2015-08-04 at 20 08 03](https://cloud.githubusercontent.com/assets/13522800/9061760/d3e986f6-3af8-11e5-8fd9-5d58fa77f897.png)
![screen shot 2015-08-04 at 20 22 19](https://cloud.githubusercontent.com/assets/13522800/9061761/d42af172-3af8-11e5-9eb4-dbba8176d700.png)
[...]

PR #494, opened 2015-08-06
Sinking Prediction Stamp and Several Fixes

ReiFan49:
* PvP with battle prediction
  * tested with PvP, and might make my kanmusu traumatized for having 3 consecutive battkes knocked out (as explained in #493)
  image
  *(positioning not tested again, the captured one still buggy, attempted to fix in ternary and positioning commits)*
* need to ensure the reflected position of experience gained during sortie is ok, #473 
* added explanation text for HQ Exp State as stated on #472 
* updated equipment counter for every successive craft

however i wondered, on some screenshots wandering around the issues,
235/100 is considered as wrong thing right? (i wonder how to fix that ._.)


bamboo3250:
@dragonjet I think @ReiFan49 pushed this commit to his master. Don't know if it's possible to merge it or not.

PR #509, opened 2015-08-07
zhCN update and fix for #510

Diablohu:
details on fix for #510

added a check for chrome.cookies for both `Play via API Link` and `Refresh API Link` function

### DIFF
--- a/src/data/translations/scn/terms.json
+++ b/src/data/translations/scn/terms.json
@@ -86,7 +86,8 @@
 	"SettingsConstructionFaces" : "剧透建造结果",
 	"SettingsCraftBox" : "剧透开发结果",
 	"SettingsCompass" : "剧透航向",
-	"SettingsBattleData" : "战斗模式",
+	"SettingsBattleData" : "战斗预测",
+	"SettingsBattleStamp" : "战斗后状态提示",
 	"SettingsBossComp" : "Boss配置",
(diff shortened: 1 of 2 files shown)


ReiFan49:
ah i see... (sounds like chrome keeps silent updating huh ....)

Crazyteitoku:
build/tmp/pages/popup/popup.js:

41 |                go_play_via_api()
                                         ^ Missing semicolon.
65 |                go_refresh_api()
                                         ^ Missing semicolon.

2 errors in 64 files

Diablohu:
semicolons added

Crazyteitoku:
Ah I was waiting for the report from Travis CI and now it shows up as checks have passed. :+1: Not really allowed to do merges, etc so I'll let @dragonjet handle it.

ethrundr:
ah those semicolons.....
so it's why I'm seeing the failed test from Travis before :laughing:

dragonjet:
wait, I think this is a matter of permissions.

I'm checking the [chrome API](https://developer.chrome.com/extensions/cookies) and there's no notice of removal on any channel.

will merge this though and edit later, since I forgot that "Play via DMM" also has a middleman page, so the `cookies.set` should be on the `\src\pages\game\web.js` not on `popup.js` :laughing:

PR #514, opened 2015-08-08
Take support expedition damage into account for battle prediction

### LINKED ISSUE #485: Ships sunk or damaged by support expedition are not marked as sunk by battle prediction
I noticed this as I was doing 5-5 that ships that were getting damaged by fire support from my support expedition were not getting marked as sunk unless they got over-killed by another ship for their full hp.

Javran:
Fix #485 
Now it should work for `api_supoort_flag = 0` (without support) or `api_supoort_flag = 2` (with hougeki support)
It should also work fine for `1` and `3` if the data format has not yet been changed (see discussion #485 for detail)
I haven't unlock world 5 but it works on data provided by @NicolaosGreene

Some bug for expedition evalutation is also fixed, but this part has not yet been used by KC3 Kai anyway.


dragonjet:
:+1:

PR #522, opened 2015-08-08
brick party (several fix)

ReiFan49:
"brick" is not "brick"
- night battle nowhps = update current HP at the start of night battle
  - morale deduction by 3 if initiated, and 1 if node basis
- give a warning/explanatory while **waiting the panel to be loaded**
- `didFyee()`
  - `didFyee`
  - some generalization applied to Fleet.js

(`spent all-night together` with @ethrundr lately)


ethrundr:
it was tiring seeing him trimming the codes......

ReiFan49:
it's called **`compressing`**

PR #532, opened 2015-08-09
Tooltips for natsuiro

McDohl:


dragonjet:
lol wait conflict :laughing: 
me, @bamboo3250 and you all made a fix for the `clearBattleData` XD
it's fixed as of latest commit as well.
maybe you can separate the tooltip and showcase edit, so we can pull only those?

McDohl:
okay

PR #568, opened 2015-08-10
Minor update

rephira:
Found some errors on ships.json and had them attended.
The error on Akitsumaru's Japanese name might be global. Needs checks (was missing the last kanji character)
Also added the new class 'Oiler' ('급유함' for the Korean version) at the very bottom of stype.json


rephira:
There was a string seemingly intended for Akitsumaru for 36 that later formed a duplicate. I got a Travis error so I ended up removing that line instead.

PR #598, opened 2015-08-12
[Fix #596] Uncaught TypeError: Cannot read property '0' of undefined

### LINKED ISSUE #596: Uncaught TypeError: Cannot read property '0' of undefined
Happened in 31-4, node E and G (yasen nodes):

31-4-E:
screen shot 2015-08-12 at 21 07 16

31-4-G:
screen shot 2015-08-12 at 21 08 32

The battle prediction is not appearing for both party. Hopefully this issue is not a duplicate, since after short glance on newest opened issue, I cannot find the similar one.

I just updated my KC3kai to this commit: 6274c95

bamboo3250:
Fix #596
@ethrundr After this gets emerged, could you test it on yasen node for me?


ethrundr:
@bamboo3250 sure, I'll help you try this :smile:

PR #600, opened 2015-08-12
Testrun Marriage Ring

ReiFan49:
- TODO: replace marriage ring picture with the correct one (i cancelled maimai again today)
- leave some _hidden_ option on ships marriage list to be tested further (you can check it by switching from `50*0` to `50*1`
- API Link Concealment (CSS trick, because lazy `_(:3/|_)=3`)
image
- Added translation support for API Leave Page (for whole page, later)
- About database upgrade:
  + it shows the change of the database in order to ascend to a newer version
  - use `ch` in order to replace the current table structure with the new one
    - replacing whole database is allowed (i'll give the "Actual Structure", which previously used)
  - use `rm` in order to remove the table
  - `vr` must be included for every new change proposal
  - `up` function is not so important, unless it's required to convert some field, or automate some values


dragonjet:
ayy xD
https://travis-ci.org/dragonjet/KC3Kai/builds/75276935

ReiFan49:
pushing it right now,
the current error is cool lol

"no anonymous function within a hardcoded loop"
i mean, really.

PR #625, opened 2015-08-13
Nodal and Battle Patch

ReiFan49:
note that SS.png and kekkon.png should be fixed right away ._.)
this one just a placeholder


ethrundr:
lol that "Travis-kun" commit :laughing:

PR #646, opened 2015-08-15
fix FCF flag reset when going to next node

bamboo3250:


dragonjet:
oh gods thank you! that `didFlee` reset was shitting me I can't find the source which code was reverting it to `false` XD

bamboo3250:
Ikr, it also took me quite a lot of time to trace the code and the sequence of API calls. The code looked no flaw to me at the first glance anyway :joy:

PR #674, opened 2015-08-18
[Feature #654] Add Night battle attack power column in Ship List in Strategy Room

### LINKED ISSUE #654: Add Night battle attack power column in Ship List in Strategy Room
In event maps, one of our big concerns is which ship will deal the most damage in night battle. Therefore, there is a need for sorting ships by night attack power, which is the sum of fire power and torpedo stats (with or without equip stats).

bamboo3250:
resolved #654


dragonjet:
i think the yasen icon wasn't included xD

bamboo3250:
oh, my bad xD

PR #733, opened 2015-08-26
update display when bucket used on docked ship

bcc32:
I hope I didn't get any off-by-one errors with the indices; I checked PlayerManager.js but I'm not entirely sure.


dragonjet:
pulling and let's see how it goes xD

though iirc, `PlayerManager.repairShips` is not based on ndock index. Ships get added there regardless of which dock they are sent to.

bcc32:
That might be a bit tough since it looks like speedchange contains the ndock index but not the ship id.

PR #747, opened 2015-09-02
Combine Normal and Event Sortie

ReiFan49:
- fixed the event tab world 0 selection bug
- only treat two-digit world as event world


ReiFan49:
i wonder where is travis-kun..

PR #825, opened 2015-09-11
Request!

Agc0de6:
@dragonjet What do you think about giving translators rights to translate web.html and api.html?
I think that english explanation how 2 turn on KC3 and then choosing a player's home language is a little bad.


ethrundr:
@dragonjet

ReiFan49:
umm i remember @Yukinyaa have said about this actually,
still haven't implement this, sorry

dragonjet:
ah yes, the end goal is to make everything translatable :book: 
we might need an issue entry for this just so it wouldn't be forgotten

Agc0de6:
@dragonjet No prob!

PR #829, opened 2015-09-11
Use translation repository as source files

dragonjet:
This is for v20 next week

Do not pull tonight will need the v19.2 copy without this.

#### What changed
* No more `src/data/translations/` directory
* The directory for translation files is now `src/data/lang/`
* `src/data/lang/` is a submodule pointing to the [translation repo](https://github.com/KC3Kai/kc3-translations)
  * this means any update to that repo can be pulled to this directory
  * it has its own travis-kun which only validates json files
  * I will be in charge of updating the submodule to point to the latest translation commit

#### Travis
Do not mind the build error, travis is unhappy with the current submodules setup. It just needs a single line on `travis.yml` which should have been included on the branches, but weren't. This `dev-translation` branch has been created before the YML fix.


nicky008:
Unhappy Travis-kun is unhappy Travis-kun... :v

ethrundr:
We'll gonna see lot of anger from Travis-kun from now on....

Issue #831, opened 2015-09-11
Any reason to represent needed XP to next HQ Lv in negative number ?


nekoworkshop:
![image](https://cloud.githubusercontent.com/assets/1346257/9818528/17392b46-5877-11e5-91b6-0fb755e2f7bb.png)

To me it's kind of unintuitive. If this behavior is not intended I may fix it.

bcc32:
I think it's pretty common to use negative numbers to refer to stuff remaining.  A lot of video players I know report current time in the video as positive and remaining time as negative.

nicky008:
As for I know now, unlike the main HQ's level above there is no any clickable choices in representing that needed EXP yet.

PS. I personally think the `jp` version text of `Next` level is a bit too long?

nekoworkshop:
In this case a negative number is representing the "*relative position*" instead of "*distance*" (which can never be negative). Just like I'll never say "There is still -6 minutes to go." as well as "-25445 exps to next level."

It really depends on whether you treate this "next" thing as "my current exp gauge *positon* relative to a full gauge"  or "*how many* exp I need to get to reach next value"


@nicky008 I'm fixing those over-size translation right now. Didn't expect we have such cramped text area...

bcc32:
I think it is fine to remove the negative sign if the label "Next" is there.  That would make more sense than having both "Next" and the negative sign at the same time.

dragonjet:
hm this is first used on ship exps, under the level on the fleet panel. instead of putting `1000` which would usually mean 1000 "current" exp, it is then made negative `-1000` to indicate 1000 remaining.

On older versions this was `Next: 1000`, a positive number becuase there was something to indicate that it was the remaining exp until next level, but because of space concerns, "Next" was removed, and was replaced by just the negative sign

nekoworkshop:
So now that we have the "Next" back, shall we go back using positive numbers ?

ReiFan49:
sounds like yes, both timer and value required will stay unsigned right?
wait for me >w<

nekoworkshop:
Let the Chief Maid do her job :3

nicky008:
@ReiFan49 It's not 2AM yet so don't hurry up like yesterday...
![capture](https://cloud.githubusercontent.com/assets/10517990/9820526/6ea7a6dc-58ef-11e5-85ee-87aa027a94e5.PNG)

ReiFan49:
somewhere on the japanese dictionary writes this..

[...]

PR #834, opened 2015-09-11
[#831] Untested -- Attempts to remove minus sign on remaining time/ex…

ReiFan49:
i just... oh forget it.

again >->o


nekoworkshop:
`KC3SortieManager.hqExpGained * (ConfigManager.hqExpDetail==1?-1:1)`

Always something new to learn everyday :3

ReiFan49:
if i remember, hqExpDetail with value 1 is "remaining", so a minus would be a good thing to be put (previously) :3

dragonjet:
I was like, :open_mouth: 

ss 2015-09-12 at 07 58 05

ReiFan49:
a mistake that summons another branch _(:3

Issue #754, opened 2015-09-04
FeatureRequest export ship config to kancolle-calc.net DeckBuilder format


c933103:
The tools' URL: http://www.kancolle-calc.net/deckbuilder.html . The data format provided by this tool is also supported by other softwares like https://github.com/Diablohu/WhoCallsTheFleet/blob/master/README.md

nekoworkshop:
@dragonjet If nobody is working on this feature at the moment you may assign me to it. I'm still familiarizing with the codebase and this feature could be a good starting point.

dragonjet:
ah sure, I added you to the Developers team which gives you write access to the main repo. It is suggested though to do feature branch :+1:

dragonjet:
wow it works XD
anything else needed for this feature?

![ss 2015-09-12 at 01 55 32](https://cloud.githubusercontent.com/assets/557312/9830151/fac9db74-5955-11e5-9258-79e348b7ef2f.jpg)

nekoworkshop:
They've got a nice set of calculators. I'm thinking about incorporate at least the damage calculator into KC3. That could come in handy in the following events by giving us a general idea of how much damage we can expect from each of our girls.

(Sorry for the weird English translation. I never played Kancolle in English nor do I know all the jargons in English :3)
![image](https://cloud.githubusercontent.com/assets/1346257/9830220/92050db4-58f3-11e5-842b-058b68613932.png)


The only thing needed to be done about the DeckBuilder exporter is to find a good place for it. I put it in the front panel just for testing (everyone on dev build will see it XD). We might want to move it to somewhere else but currently I have no idea...

c933103:
Have aircraft proficiency included in the export data?

nekoworkshop:
@c933103 The deck data generated by the official deck builder doesn't include aircraft proficiency ( it does include an "rf" field for each aircraft though. This field is used as star numbers by other equipment but has no effect on aircraft in the deck builder.).

If I missed anything please feel free to enlighten me.

PR #843, opened 2015-09-12
[Feature] kancolle-calc.net Deckbuilder export feature

nekoworkshop:
Feature requested in #754. Click the button and try it out yourself. ( I know the icon doesn't look pretty. Sorry...)

image

Our panel is running short of space. Please share your idea about where we should put it. Much appreciated.


nekoworkshop:
I kept the codes in the wikia export page just in case we need to move the feature back there. I'm not sure whether we should  put it in the Strategy room or front panel.

Issue #690, opened 2015-08-20
[Feature request] kancolle-db.net integration


chronomeister:
So I'm noticing more people being interested in kcdb. The people there wrote their own app for KCV and others to proxy chain through and deliver the info to kcdb. The main part of their module can be found here: https://github.com/about518/kanColleDbPost/blob/master/KanColleDbPost/Form1.cs

But the code in particular is this:
```
// oSession is, as you can guess, the original request/response from the kc servers.
string token = textBox2.Text; // token given by kcdb
string agent = ""; //currently unused
string url = oSession.fullUrl;
string requestBody = HttpUtility.HtmlDecode(oSession.GetRequestBodyAsString());
requestBody = Regex.Replace(requestBody, @"&api(_|%5F)token=[0-9a-f]+|api(_|%5F)token=[0-9a-f]+&?", "");
string responseBody = oSession.GetResponseBodyAsString();
responseBody.Replace("svdata=", "");

try
{
   WebRequest req = WebRequest.Create("http://api.kancolle-db.net/2/");
   req.Method = "POST";
   req.ContentType = "application/x-www-form-urlencoded";

   System.Text.Encoding enc = System.Text.Encoding.GetEncoding("utf-8");
   string postdata =
        "token=" + HttpUtility.UrlEncode(token) + "&"
      + "agent=" + HttpUtility.UrlEncode(agent) + "&"
      + "url=" + HttpUtility.UrlEncode(url) + "&"
      + "requestbody=" + HttpUtility.UrlEncode(requestBody) + "&"
      + "responsebody=" + HttpUtility.UrlEncode(responseBody);
   // next part is just how you do web requests in C#
   byte[] postDataBytes = System.Text.Encoding.ASCII.GetBytes(postdata);
   req.ContentLength = postDataBytes.Length;

   Stream reqStream = req.GetRequestStream();
   reqStream.Write(postDataBytes, 0, postDataBytes.Length);
   reqStream.Close();
}
```

Should be pretty easy to follow. token is kcdb's token provided when you sign in through twitter. Gets rid of any personally ident api info. Gets rid of the svdata header. Then sends the whole json to them for them to process. Otherwise just change from C# to javascript.

Doesn't require much in the way of presentation; On setting page just make a label (with a link to either website or explanation), a checkbox to confirm/deny sending data and then a textbox to put your token/key. Everything else is backend.

nicky008:
I'm also a fan in KCDB too. :raising_hand: 

[...]

PR #895, opened 2015-09-24
Kancolle db submission

nekoworkshop:
Feature requested in #690 

You may obtain a access key at http://kancolle-db.net/login.php ( A twitter account is needed ) . Navigate to the setting page, enable the feature, paste the key and you're done.

You may check the collected data  at the above page.

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -63,6 +63,24 @@
 		]
 	},
 	{
+		"section": "SettingsDBSubmission",
+		"help" : "",
+		"contents" :[
+			{
+				"id" : "info_DBSubmission",
+				"name" : "SettingsEnableDBSubmission",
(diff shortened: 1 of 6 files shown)


nekoworkshop:
About the error message.

I couldn't get rid of the  `'Access-Control-Allow-Origin'` error from my side so after hours of struggling I contacted the author of KancolleDB and he kindly added the required response header.

There do exist several workarounds (see https://github.com/KC3Kai/KC3Kai/issues/690#issuecomment-140247023) but the all of them can only deal with the case when  `'Access-Control-Allow-Origin'`exist but doesn't satisfy the security policy specified in the manifest file. There is nothing we can do if there is no  `'Access-Control-Allow-Origin'` at all.

PR #910, opened 2015-09-30
Game page translations

dragonjet:
Don't merge yet until reviewed


kololz:
I will be handling that as soon as possible

PR #918, opened 2015-10-02
Allows ECMAScript6

ReiFan49:
leaving to other for review this syntax expansion to ECMAScript6

http://www.zev23.com/2014/04/use-gruntjs-to-uglify-es6-application.html

PS: i forgot to make a branch inside :sweat_smile:
PPS: i only use `copy`, `jsonhint`, `jshint`, `clean` for local travis check

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,4 +1,5 @@
 {
     "newcap": false,
-    "laxbreak": true
+    "laxbreak": true,
+    "esnext": true
 }
--- a/src/library/modules/Kcsapi.js
+++ b/src/library/modules/Kcsapi.js
@@ -924,66 +924,46 @@ Previously known as "Reactor"
 		var rankPt = getRank(data.api_win_rank);
 		if(rankPt==5 && KC3SortieManager.currentNode().allyNoDamage) rankPt++;
 		if(!isPvP) {
-			while(rankPt>=3) {
-				switch(rankPt) {
-					case 6: // PERFECT S
-					case 5: // S
-						KC3QuestManager.get(214).increment(3); // Bw1: 4th requirement: 6 S ranks (index:3)
-						
-						if(KC3SortieManager.currentNode().isBoss()) {
-							switch(true) {
-								case KC3SortieManager.isSortieAt(5,2):
-									KC3QuestManager.get(243).increment(); // Bw9: Sortie to [W5-2] and S-rank the boss node 2 times
-									break;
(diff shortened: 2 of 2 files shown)


dragonjet:
I can't read it anymore. seems to be obfuscated now :open_mouth: 

to be honest, one more goal I forgot to mention is readability, for other devs. one of our edge is that we use plain javascript (not coffeescript or anything else), something that anyone can read, and that's probably the reason why we got so many contributors.

But if we start converting stuff to ninja, I'm afraid even I can't fix stuff when a bug occurs or an API change or something :sob:

bcc32:
As it stands right now though, I think the current code is pretty confusing already.  I like the idea of separating out the data for the quests into a table... perhaps it would be helpful to add some comments as to what each entry is, i.e., 

    [quest ID#, counter to increment, specific map requirement, boss node required].

Or maybe this information should belong in quests.json, somehow?

ReiFan49:
@dragonjet i thought you were talking about ES6 thing :sweat_smile:,
yeah, forgot to give the explanation for the data, 
kinda forgot to "say propose" like previous data proposal :grinning: 

@bcc32 leaving it in quests.json means translation right? and it might be a bit redundant i guess to make it dupe over language/quests.json itself (like `unlock` and `tracking`) leaving it "hardcoded" is better because the value updates as more quests to come

the quest requirement structure is like this
```
KeyLabel   DataType    Description
QuestID    Integer
IncrIdx    Integer     leave it to `zero` or `undefined`
                       to set it to the `index zero`
Map        Integer[2]  it can consists up to 2 integers.
                       no data means all map,
                       single data means all map within the specified world,
                       paired data means only happens on specific map
HornChk    Boolean     is this only happens in horned nodes, or every nodes?
```

dragonjet:
@ReiFan49 ah no just talking about this block of code XD
but reading it outside of github code, I can still follow. whoo Atom editor~ :sweat_smile: 

you could probably add on the js comments what are those four elements per quest so anyone who tries to fix something can easily understand it

@bcc32 i think that array can stay on the file, since that code block only applies to sortie. If we try to separate requirements to a json file and want it standardized, it's difficult to apply to single counting like expedition, resupply and stuff, in the end, it still does `foreach` on `qLog` after all xD
```
.forEach((x)=>{
	qLog(x[0]).increment(x[1]);
});
```

ethrundr:
accidentally pressed closed, and just realize....

dragonjet:
was mentioned to be re-applied at #1104

Issue #720, opened 2015-08-23
[Ask] What happened to the heartlocks in strategy room?


LostLogia4:
Seems like it was gone in favor for night battle FP in latest webstore version...

bamboo3250:
Yes, there is not enough space for the heartlock so it's pushed to outside of the table xD

NicolaosGreene:
Is there a reason we can't widen the table further?

dragonjet:
we're limited to 800px, the same width as the game

NicolaosGreene:
Might we want to make the stats toggle-able then so that we can see the ones that we are interested in then?

NicolaosGreene:
Another option would be a scrollbar for when it is overflowing as well.

LostLogia4:
How about a toggle option between firepower + torpedo and night battle firepower?

dragonjet:
it would be nice if there's a column selector where you can enable/disable columns, but I guess that's too much for a hard-coded table like ours. At these times we'd want to use JS plugins for table, but I'm always concerned about filesize of the extension in the webstore

bamboo3250:
I would go down to "column selector" route. There will be some must-have columns and some optional columns for the column selectors. Hopefully I can finish it at the end of this week. I would be busy since new semester has come though.

dragonjet:
:+1:

mdms-scott:
@dragonjet Out of curiosity, is the 800px limitation self imposed or some kind of actual restriction? (Modifying the css inline to increase the width of the table seems to show things fine)

dragonjet:
@mdms-scott it's self-imposed, knowing that the game width is `800px` and most people would open the strategy room in the same window as the game

mdms-scott:
@dragonjet Ahh, makes sense, although I would kind of assume that most people using the strategy room will also have the KC3 panel open and thus would have a wider window, but the column selector route makes it kind of unnecessary.  It occurs to me that the strategy room could be reactive I suppose?

dragonjet:
hm I though about this in the first redesign i think around v10, but sticked to fixed 800px, for the purpose of this app, to make it look how we want it to, and is exactly the same for all users.

Other than this app, I use responsive layouts in all websites I make, which easily supports mobile phones and tablets, and desktop alike. However, this is not a website and will primarily be viewed from a desktop, which have a minimum of 800px for the game, so with both points imho we should stay fixed width.

[...]

PR #920, opened 2015-10-05
Adding a heartlock filter

Javran:
Adding a heartlock filter in ship list page.

2015-10-05_153700_3360x1080_scrot

could be one solution to #720

### DIFF
--- a/src/pages/strategy/tabs/ships/ships.html
+++ b/src/pages/strategy/tabs/ships/ships.html
@@ -45,6 +45,12 @@
 			<dd class="hover fleet_yes">Yes</dd>
 			<dd class="hover fleet_no">No</dd>
 		</dl>
+		<dl>
+			<dt class="massLabel">Heartlock:</dt>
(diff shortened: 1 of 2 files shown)


ReiFan49:
```
(self.heartlockFilter === 0
  || (self.heartlockFilter === 1 && thisShip.locked === 1)
  || (self.heartlockFilter === 2 && thisShip.locked === 0)
)
```
not sure if that is a bit strict comparing than the previous one... :P 
```
(
  ((self.heartlockFilter || 1) == 1 && thisShip.locked)
  || ((self.heartlockFilter || 2) == 2 && !thisShip.locked)
)
```

Javran:
personally I prefer the `===` comparison, as I think it's more readable and JS is a language that you can't be too careful xD

also for bitwise operator you need `|` instead of `||`:

```javascript
> 1 || 4
1
> 1 | 4
5
```

dragonjet:
agree with @Javran, I only use `||` for `undefined`

PR #925, opened 2015-10-07
Strategy room: ship docking time

Javran:
Modified from ship list, shows non-full HP ships with docking time / facility repair time (Akashi time)

The screenshot should be straightforward:

2015-10-07_152406_3360x1080_scrot

### DIFF
--- a/src/library/managers/PlayerManager.js
+++ b/src/library/managers/PlayerManager.js
@@ -59,6 +59,7 @@ Does not include Ships and Gears which are managed by other Managers
 		setRepairDocks :function( data ){
 			var lastRepair = this.repairShips.map(function(x){return x;}); // clone
 			this.repairShips = [];
+			var dockingShips = [];
 			var self = this;
 			$.each(data, function(ctr, ndock){
 				if(lastRepair[ndock.api_id] != ndock.api_ship_id) { // check if not in the list (repaired)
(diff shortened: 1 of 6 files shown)


dragonjet:
oh very nicely done :+1:
I never even though of this XD

dragonjet:
can I edit the page in the branch? I'd like to add HP bars and some other minor ui tweaks

Javran:
@dragonjet feel free to edit :)

Javran:
@dragonjet also there are few things I haven't done, but I won't have time in these few days to take care of them, feel free to take over or comment:

* Refresh the page when switching to it, so remaining docking time will be up-to-date, for now the page refreshes when sorting, but not when switching from other strategy room pages

* `PlayerManager` now adds a field `dockingShips` to localStorage and `docking.js` reads it directly, I guess a better way would be to have some API in `PlayerManager` for reading the data.

 * [shouha / chuuha / taiha status is calculated individually](
https://github.com/KC3Kai/KC3Kai/blob/dev-docking-page/src/pages/strategy/tabs/docking/docking.js#L178-L193) instead of using existing methods in ship object. would be nice to move this part into ship object so other modules can use.

* Mark ships on expedition (like those docking ships but with another color)

* Show some alternative texts when no ship needs repair

* Help topics

Crazyteitoku:
Marking this for documentation.

PR #945, opened 2015-10-12
Adding stat sorters on equipment page

Javran:
Added few buttons on equipment list, the list now can be sorted by stats

2015-10-12_101333_3840x1080_scrot

and these buttons are changed dynamically depending on equipment types and equipments available.

2015-10-12_101428_3840x1080_scrot


Crazyteitoku:
Marking this for documentation.

Issue #952, opened 2015-10-17
poi-statistics DB submission


Javran:
similar to kancolle-db integration,
send item crafting / ship drop / ship construction results to [poi-statistics](http://db.kcwiki.moe/).

self-assigned, probably make some progress this weekend.

c933103:
see https://github.com/poooi/poi/issues/460

Javran:
@c933103 thanks for the info :)

implemented `create_item`, `create_ship` and `dropship`.
await further discussion on https://github.com/poooi/poi-server/issues/1 to continue:

- [ ] server side CORS support
- [ ] set `origin` field instead of setting User-Agent (which is impossible for a chrome extension).

Javran:
done.

PR #955, opened 2015-10-18
poi-statistics db submission

Javran:
done #952


Javran:
@KC3Kai/translators 2 more terms to be translated, see:

https://github.com/KC3Kai/kc3-translations/commit/bef39ea09bd43ae4da53614184ccae9f31d59070

PR #956, opened 2015-10-18
List all equipments on equipment page

Javran:
Adding a new button to equipment list, which lists all equipments, the sorter added previously can therefore used to sort all equipments by stat regardless of type.

2015-10-18_111856_3840x1080_scrot

an icon for it is still missing, but I have no idea which one should I use.


ReiFan49:
try to use standard gear icon with "ALL" text above the gear icon...

Javran:
I didn't find a good icon to use, or just using text would be fine?

dragonjet:
since we cannot use any of the existing equipment icon to represent all equipment, text would probably sufffice

Crazyteitoku:
We could probably use some of the in-game assets such the ones they have featured in the drop-down menu:


































For reference these were from commonAssets.swf.

Issue #978, opened 2015-10-28
few more tasks for expedition tab


Javran:
keeping a TODO list:
- [x] remember user's last selection
  fleet 2,3,4 could have their own config about what expedition to go and whether it should be calculated with great success, triggered when switching between fleets and expedition tab is visible
- [x] colorize expeditions by availablity
  - use different colors for expeditions that current fleet can / cannot do
  - also `api_get_member/mission` could provide some clue about what expeditions are available to current player (might need some testers since I've unlocked all expeds)
- [ ] ~~battle tab can be switched out for exped tab when back to port, likewise when going sortie exped tab can be switched out for battle tab~~
  - in case someone still want to see that tab, shrink header width (change text to "...") to make room for other available tabs instead of hiding it.
- [x] fuel & ammo consumption is changed for married ships, not taken into account for now (need testers since I don't have any of them)
- [x] some more enhancement for making the tab smarter
  - switch to expedition tab when at the expedition selection screen (when `api_get_member/mission` is received)
  - show the first fleet not running an expedition
  - after sending out a fleet, if there is another available fleet, show it in the expedition tab
- [x] provide a setting for totally disable this tab
- [x] if all other fleets are sent out, focus returns to main fleet

@nekoworkshop feel free to add things you've planned but not yet implemented.

self-assigned for now.

NicolaosGreene:
Is there a plan to make focus return to the main fleet after you have sent out all the expeditions?

Javran:
@NicolaosGreene good idea, will do.

Javran:
pretty much done, self-unassigned.
not sure how to hide / show tab headers properly: seems like the width is hard-coded in natsuiro.css but I want to adjust it dynamically (see OP)
so here I leave the last task unfinished, any one interested can take over, feel free to ask for details xD

dragonjet:
I want to see this but maintenance, can't open devtools :V

dragonjet:
hm not sure I can follow this XD
>in case someone still want to see that tab, shrink header width (change text to "...") to make room for other available tabs instead of hiding it.

Javran:
@dragonjet now the tab header looks like:

> Basic | Battle | Exped | Activity

I want it to be:

> Basic | Battle | ... | Activity

when going sortie, and:

> Basic | ... | Expedition | Activity

when going back to port.

Javran:
[...]

PR #983, opened 2015-10-29
Expedition tab

Javran:
See checked item in #978 for progress.

Implemented:

* remember user's last selection for each fleet
* colorize expeditions by availablity (feel free to edit CSS colors because I'm not good at this)
* under some conditions expedtion tab will be activated automatically:
    * if there are fleets available for expeditions, switching to expedition selection screen will activate expedition tab
    * after sending a fleet to expedition, if there are still fleets available, that fleet will be selected automatically

### DIFF
--- a/src/assets/js/global.js
+++ b/src/assets/js/global.js
@@ -273,6 +273,6 @@ Array.prototype.equals = function (array) {
         }           
     }       
     return true;
-}
+};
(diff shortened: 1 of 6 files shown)


dragonjet:
oh alright, i will push the image to expedition-tab and pull request again

PR #1092, opened 2015-12-24
modification to leveling tab

reynn:
set image title to ship name + rosterid this can make it easier for newer players or ships that some of us don't recognize easily to be selected. Also increase max level to 155 per 12/8 kancolle release.


dragonjet:
thanks :+1:

dragonjet:
added @reynn to dev list, you can give up to 5 links for the credits section

Issue #541, opened 2015-08-09
Bw8 is not marking as completed when the conditions are fufilled


NicolaosGreene:
The completed conditions for Bw8 is to defeat (B-rank minimum) a boss in World 4-4. However, currently the quest tracker does not reflect this and did not mark as complete when I got an A-rank by scoring kills on ships 2 through 5.

NicolaosGreene:
this doesn't even mark as complete if you S-rank rank the node as an update to this

NicolaosGreene:
Appears to be working correctly now closing

PR #1117, opened 2016-02-01
fix quest Bw8 tracking (4-4)

Javran:
Hopefully this fixes Bw8 tracking ( #541 ). I've completed my Bw8 so I might need verification on this @KC3Kai/testers


Javran:
hit enter accidentally XD

I think the bug is caused by this part of code: https://github.com/KC3Kai/KC3Kai/blob/master/src/library/modules/Kcsapi.js#L1542-L1547

if `KC3SortieManager.isSortieAt(4,4)` is true, then `KC3SortieManager.isSortieAt(4)` must be true, so it shortcuts the tracking code for Bw8 (quest id 242)

PR #1127, opened 2016-02-07
"Other forms" in Ship Library

Javran:
I have implemented a module called "RemodelDb", that computes everything related to remodeling (cost, required level, etc.) from `api_start2` data. devs can take a look.

As a byproduct of this. I added "other forms" in ship library. when a ship has forms other than herself and her next remodel, "Other forms" will appear. may need some tweak on UI afterwards xD.

2016-02-06_223726_3840x1080_scrot

Note that for this patch to work you need to start the game so RemodelDb can receive `api_start2`. (you only need to do this once)

### DIFF
--- a/src/library/modules/RemodelDb.js
+++ b/src/library/modules/RemodelDb.js
@@ -1,0 +1,177 @@
+/* RemodelDb.js
+
+   Everything related to ship remodels
+
+*/
+(function() {
+    "use strict";
+
+    window.RemodelDb = {
+        _db: null,
+        // NOTE: masterData on init is optional
+        // as long as pre-processed data has been saved in localStorage
+        init: function(masterData) {
+            // load stored db if any
+            if (typeof localStorage.remodelDb !== 'undefined')
+                this._db = JSON.parse( localStorage.remodelDb );
+
+            if (masterData && this.requireUpdate(masterData)) {
+                try {
+                    var db = this.mkDb(masterData);
+                    localStorage.remodelDb = JSON.stringify(db);
+                    this._db = db;
+                    console.log("RemodelDb: database updated");
+                } catch (e) {
+                    console.error("RemodelDb:",e);
+                }
+            } else {
+                console.log("RemodelDb: no update required");
+            }
+
+            if (! this._db) {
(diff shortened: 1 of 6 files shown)


src/library/modules/RemodelDb.js
```diff
@@ -0,0 +1,202 @@
...
+
+            // patch devmat info, can't find it in api_start2
+            console.assert( remodelInfo[461].ship_id_to === 466 );
+            remodelInfo[461].devmat = 15;
+            console.assert( remodelInfo[462].ship_id_to === 467 );
+            remodelInfo[462].devmat = 15;
+            console.assert( remodelInfo[466].ship_id_to === 461 );
+            remodelInfo[461].devmat = 10;
+            console.assert( remodelInfo[461].ship_id_to === 466 );
+            remodelInfo[462].devmat = 10;
```
ReiFan49: i feel that there's something is not right on here....

dragonjet: I didn't even notice remodelling needs devmats? Of all years I play

Javran:
@dragonjet @ReiFan49 I've sync-ed latest RemodelDb.js into this branch (to eliminate hard-coded rules), and somehow your comments are gone, so I'll just reply here: devmat consumptions are previously hard-coded for cranes, but it turns out 74EO's document says there are rules about remodel devmat consumption:

https://github.com/andanteyk/ElectronicObserver/blob/d28e1d904521ea60b25c15010d06f130ae36dc62/ElectronicObserver/Other/Information/kcmemo.md#%E6%94%B9%E8%A3%85%E6%99%82%E3%81%AB%E5%BF%85%E8%A6%81%E3%81%AA%E9%96%8B%E7%99%BA%E8%B3%87%E6%9D%90

dragonjet:
oh good find :+1:

ReiFan49:
umm actually, you can reply on the commit,
it's not lost, but you just need to refer which commit though (looking through the commit history is the only thing that possible)

dragonjet:
I clicked on the "show outdated diff" on this thread :laughing:

PR #1144, opened 2016-02-12
Handle ld_airbattle for combined fleet

Javran:
I haven't seen any data sample so this patch is just guessing. (I'm 90% confident)
need volunteers on E-3 (I haven't got there yet).


mikenzb:
Tested on E3, it works. someone with merge powers should merge this into master.

ReiFan49:
so basically it's the same thing as the other battle right?

mikenzb:
Yeah it's the same as the AA node on E2 but the only difference here is it's in a combined fleet vs the landbase

PR #1145, opened 2016-02-12
unify shipIcon and abyssIcon

Javran:
making KC3Meta.shipIcon and KC3Meta.abyssIcon the same


dragonjet:
oh nice, I was previously thinking when I first saw, we'll need to edit alot of scripts that use them but assigning it to the function names works well too. good work :+1:

Issue #1170, opened 2016-02-28
rework on quest tracking


Javran:
Despite #1117, there are still problems: if user choose to complete their Bw6 (Defeat 12 bosses in the World 4) on 4-4, we will have problem tracking it.

I feel the code related to quest tracking:

https://github.com/KC3Kai/KC3Kai/blob/5787ec93b497dd7bbdd294d442c16f2852609cec/src/library/modules/Kcsapi.js#L1528-L1607

is a bit hard to understand, and using straightforward if-statements will be better for both understanding and maintenance.

I'd like to ask if there are any advantages of using current structure (looping+switching)?
if not I can assign myself working on this to make it look better. (and solve problems like #541 once and for all)

dragonjet:
agreed with conversion to ifs, haven't seem this block before though

ReiFan49:
https://github.com/KC3Kai/KC3Kai/blob/dev-api/src/library/modules/Kcsapi.js#L1639-L1693

well, grouping them using switch-while combo just to make them a bit neater

Javran:
just discussed with @ReiFan49 on gitter channel. another alternative is #918 

@ReiFan49 is more in favor of switch-looping and #918 and I'd recommend #1173 but is also good with #918 (with more  comments to explain). wondering what do you guys think?

ReiFan49:
and also, this need to be solved first before merging the `dev-ledger` branch though

dragonjet:
wait, is ecma6 safe? I know we are already recommendinh Chrome 45+ but are there anything else we need to look out for? I heard something about babeljs that lets you use these safely in a bit older versions of browsers.

dragonjet:
can give sample code on how you guys are planning to apply ecma6 into the quest tracking. I'm starting to feel alright with it since as mentioned, we're already requiring chrome 45+. seems this is a blocker for #1104 which is near `merge-ready` state

Javran:
we can break the dependency, I suggest keeping the quest tracking part intact and merge rest of the changes from `dev-ledger` unless `dev-ledger` branch has been well-tested on quest tracking (reminder: 4-4 should count to Bw6 tracking).

because quest tracking is one feature players would need every day, we have to be more careful.
since it's during maintenance, we have no way to verify.

dragonjet:
this is included in the current v27 release right?

ReiFan49:
according to the dev-ledger previous state
- it's not ECMA6 anymore
[...]

PR #1173, opened 2016-02-28
Rework quest tracking

Javran:
Solve #1170. Still needs verification.

@KC3Kai/testers could you test this patch? just keep an eye on daily / weekly / monthly sortie quests, everything should work as usual. the only difference would be that now defeating 4-4 boss should also count to Bw6 (Defeat 12 bosses in horned nodes in World 4)


Javran:
Most of the quests are verified, worked as expected. I think it's ready for merge.

dragonjet:
:+1:  thanks

PR #1218, opened 2016-03-24
Fix #49 Detect devicePixelRatio for screenshot

### LINKED ISSUE #49: Screenshot issues
Screenshots don't work at all when you are playing on the actual DMM site, neither tab nor download.
It does work when you are playing via "DMM frame" or "API Link", except that it doesn't actually make a screenshot of the game itself, instead it seems to be a screenshot of the upper left corner of the browser that's also really zoomed in. 

I have the KC3 Kai panel at the horizontally at the bottom of the screen btw. Although I tried putting the panel vertically, it still just makes a zoomed in screenshot of the upper left corner.
I'm also on OSX/Mac and on the Kanoya Airfield server, if this helps.

Oh and I had that issue since the old KC3 by the way

dragonjet:
* Detects `window.devicePixelRatio` and applies it to screenshot scales
* Would detect both retina (2dppx) and other high-density (1.3dppx etc)

### DIFF
--- a/src/library/objects/Screenshot.js
+++ b/src/library/objects/Screenshot.js
@@ -1,7 +1,7 @@
 var imgurLimit = 0;
 
 function KCScreenshot(){
-	this.scale = ((ConfigManager.api_gameScale || 100) / 100);
+	this.scale = ((ConfigManager.api_gameScale || 100) / 100) * (window.devicePixelRatio || 1);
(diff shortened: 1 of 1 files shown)


dragonjet:
hm this is.. err... travis failed because of his connection issue, not our code. the branch is done, but we need to commit something again just so he can re-try building it

Issue #1003, opened 2015-11-03
Abyssal names on Battle Data


dragonjet:
> May I suggest an addition of the names of the ships during the battle screen for sorties and PVP? I've been on the forums for a while and have no idea what people mean when they refer to [Mod] KC3改Nu-Class or [Mod] KC3改Ri-Class Flagship etc, it would be quite nifty to know what the abyssals are called and what they look like. It would be really useful for PVP too if I see a ship that I like but only have a split second to grab its kanji before typing it into the wiki (I'm Chinese so I can read most of the Chinese characters).

from [st0rmbreaK on himeuta.net](http://www.himeuta.net/f39-kancolle-modules/mod-kc3ae-5962-21.html#post871438)

ReiFan49:
[Mod] KC3Kai ?

dragonjet:
ah they just call it that for all third-party apps, even kcv

wrinkledm:
The most recent update (27.6?) now includes the enemy PVP fleet weapon loads when hovering over the opposing fleet icons on the battle tab. Thanks for that update.  I agree with st0rmbreaK, it would be most helpful if including the ships name &  her level for those of us who still haven't learned all the ships names.

dragonjet:
I believe we already have this too?

PR #1229, opened 2016-03-27
Add ship name to tooltips on battle and PvP

sinsinpub:
Hint by last comment of #1003

And also:
* Fix i18n miss for plain theme
* Fix quest tracking and map without gauge


dragonjet:
:+1:

PR #1255, opened 2016-04-02
Some fixes for game updates

sinsinpub:
For now...
* Separated usage of `api_version` array on card and voice
* Supports separated subtitle fade-out timing for each languages


dragonjet:
thanks for good work :+1:

PR #1259, opened 2016-04-02
[dev-quest-track] Rework on quest tracking

### LINKED ISSUE #1170: rework on quest tracking

dragonjet:
#### Issues Addressed

Merging this commit will automatically close the following issues
- Close #1170


dragonjet:
unfinished feature that makes use of concepts that may never be implemented (packages). closing and deleting for now. there are better approaches than this.

PR #1264, opened 2016-04-03
[dev-premodel-quote-reuse] Reusing ship pre-remodel quotes

### LINKED ISSUE #1249: subtitle quotes translation support
now subtitle supports have been pushed to master branch, lots of translations are needed.
here are some ideas about adding better support for translators:

- use descriptions as line keys rather than using the number

current keys are numbers
```javascript
...
   "80": {
        "1": "I am Battleship Nagato, pleased to meet you. Leave the enemy battleships to me.",
        "25": "Born as the first battleship of the 8-8 fleet programme, the nameship of the Nagato-class battleships, Nagato. Until Yamato was commissioned, I served as the flagship of the Combined Fleet. I was called one of the 'Big 7'.",
        "2": "What is it?",
...
```

it could be easier for translators if the keys are more descriptive:
```javascript
...
   "80": {
        "Intro": "I am Battleship Nagato, pleased to meet you. Leave the enemy battleships to me.",
        "Library": "Born as the first battleship of the 8-8 fleet programme, the nameship of the Nagato-class battleships, Nagato. Until Yamato was commissioned, I served as the flagship of the Combined Fleet. I was called one of the 'Big 7'.",
        "Poke(1)": "What is it?",
...
```

the mapping between numbers and descriptions has been established in:
https://github.com/KC3Kai/KC3Kai/blob/master/src/pages/strategy/tabs/mstship/mstship.js#L12-L66

so we can either change numbers to descriptions or having some helper program to allow translators to write out "descriptive-key" version and convert to the current one.

- reuse pre-remodel ship quotes

as many ships are just reusing quotes of their pre-remodel version, in `quotes.json` we can just add quotes that are different or missing from their pre-remodel version. and we can use https://github.com/KC3Kai/KC3Kai/blob/master/src/library/modules/RemodelDb.js to figure out ship id before remodeling, so there is no need to put pre-remodel relation into quotes.json

Javran:
#### Issues Addressed
Merging this commit will automatically close the following issues
* Close #1249 

#### Update

adding RemodelDb.previousForm that does the following:

```javascript
        // return ship id of the previous form.
        // if the ship is already in original form, "false" will be returned
        // this relation is not circular - tracing from final form using "previousForm"
        // will not result in infinite loop.
```

one step towards #1249


Javran:
implemented 2nd bullet of #1249:

- if one line is missing, line from her pre-remodel form will be used (in that specific language)
- if the pre-remodel still doesn't have that line, English version will be the last resort.

I can confirm it's working in scn and is ready to merge, might need @KC3Kai/testers @KC3Kai/translators to try in other languages for good measure.

things to be verified:
- existing quotes should not be overwritten by anything else
- missing quotes are filled using ship's pre-remodel form. (e.g. If Ryuuhou Kai's Intro is not set, but Ryuuhou's is available, that line will be used)
- if pre-remodel quotes are still not available, English version will be used.

Javran:
gonna try few ships' quotes tonight. will merge this if everything works fine.

Javran:
Done all things mentioned in #1249, except that we might need a tool for translators to convert `quotes.json` from one with quote-id key one to descriptive key one.

sinsinpub:
but why we are not able to have no adapter function which can accept both 'descriptive key' and 'voice line key' instead of an external tool...?

Javran:
@sinsinpub the adapter function is already there: both are accepted for now with descriptive key having more priority. the reason for having that external tool is to make `quests.json` file itself more readable.

Javran:
@dragonjet conflicts are resolved, tested locally, ready for merge xD.

dragonjet:
ah perfect :+1: 
conflicts resolved, merging now

PR #1268, opened 2016-04-03
[dev-api-2] Re-implement API Page updates

dragonjet:
this is the latest master with the API page updates that @ReiFan49 added


sinsinpub:
Confirmed ration on ex-slot affect:
Both Rank S, 9 drums, 3 daihatsu.
First trial: 122TP
Second trial, just add 1 ration to ex-slot: 123TP

btw, why not pick TP calc to another branch... it does not depend on the API game codes, right?

dragonjet:
API page being switched out to special DMM mode soon. closing this

dragonjet:
deleting branch for now as cleanup

PR #1273, opened 2016-04-04
[dev-subtitles] Credits to subtitle contributors

dragonjet:


Javran:
also need to credit @kcwikizh and all contributors of them.
https://github.com/KC3Kai/kc3-translations/pull/106 will pull in their great works :)

PR #1275, opened 2016-04-09
[dev-dev-only-tab] Developer-only tabs

Javran:
as described in #1270 

- submenu "Dev-Only" in strategy room. only visible when `SettingsDevOnlyPages` is turned on.
- a playground tab for developers to do experiments, see details in https://github.com/KC3Kai/KC3Kai/blob/dev-dev-only-tab/src/pages/strategy/tabs/playground/README.md

PR #1277, opened 2016-04-09
Translation Previewer

Javran:
Requires #1275 so I'm making PR on `dev-dev-only-tab` instead of the master.

A dev-only page for translators, missing items are colored in red xD. 

2016-04-09_051011_3840x1080_scrot

### DIFF
--- a/src/library/modules/Translation.js
+++ b/src/library/modules/Translation.js
@@ -48,15 +48,11 @@
 			$("html").attr("lang", ConfigManager.language);
 		},
 		
-		
-		/* GET JSON
-		Used by KC3Meta.js to load json files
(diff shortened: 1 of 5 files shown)


src/library/modules/Translation.js
```diff
@@ -47,16 +47,39 @@
...
+						// should work for both arrays and objects
+						obj[k] = track(v);
+					});
+				} else {
+					return {val: obj, tag: tag};
+				}
+				return obj;
+			}
+
+			console.assert(
```
dragonjet: never knew console have assertion functions XD

dragonjet:
:+1:

PR #1293, opened 2016-04-16
Strategy room dev-only page: quotes helper

Javran:
a quick and dirty quote translation helper. feel free to tune the UI. 

2016-04-15_233752_3840x1080_scrot

left side list shows translation progress for each ships. green bar are those lines directly available, blue bars are inherited from previous remodels. gray bar are missing.

right side list shows voice lines for a specific ship. red background for missing quotes, blue background for inherited quotes, green background for those directly available.

### DIFF
--- a/src/library/modules/Master.js
+++ b/src/library/modules/Master.js
@@ -7,23 +7,23 @@ Saves and loads significant data for future use
 */
 (function(){
 	"use strict";
-	
+
 	window.KC3Master = {
 		available: false,
-		
+
 		_raw: {},
-		
+
 		init: function( raw ){
 			this.load();
-			
+
 			if(typeof raw != "undefined"){
 				return this.processRaw( raw );
 			}
-			
+
 			this.updateRemodelTable();
 			return false;
 		},
-		
+
 		/* Process raw data, fresh from API
 		-------------------------------------*/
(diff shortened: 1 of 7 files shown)


sinsinpub:
oh
* may be better to list up origin text (en/jp) comparing with TLed one
* and put some voice playing button for each row to check
* `From shipName` just need one time instead of every rows
just similar thing with: http://revise.kcwiki.moe/v2/#!/List/353 u know :smiley_cat: 

ofc I could continue tuning when I have time

Javran:
@sinsinpub sure, feel free to take over, and sorry for the messy code.

btw voice buttons are already clickable. I moved that part from mstship tab.
also `From shipName` has to appear many times, some ships inherit quotes from not just pre-models but also pre-pre models :sweat_smile:

dragonjet:
ohh looks good :+1: mark as merge ready when done

Javran:
I've done refactoring the code and think this is good enough to be merge-ready -- the idea is to make it available to translators asap, and it's supposed to be a dev-only page, so UI tuning is not a priority.

I'll leave @sinsinpub's first suggestion to future, since most jp quote translations are not available.
also quotes default to en version when missing, so I consider comparison with en quotes done.

PR #1296, opened 2016-04-16
minor fix to GearManager

Javran:
details are in code comments.

basically nothing is changed. just that when writing testcases, one should avoid putting itemId = 0 in GearManager.


dragonjet:
alright, so I checked my starter ship, and she's `1`. Assuming ships and slotitems follow the same logic, first equipment should also start at 1.

PR #1297, opened 2016-04-17
[dev-battle-prediction] UI and Fixes

### LINKED ISSUE #1292: Update damecon properly on UI
Now that battle prediction also predicts damecon usage. I think we can start to think about updating equipment info for damecon properly on UI:

* on battle result screen, if damecon is consumed, remove that from UI (I'm not sure whether this part has been implemented)
* as @sinsinpub has suggested in https://github.com/KC3Kai/KC3Kai/issues/1284#issuecomment-210549076 if battle prediction says a damecon will be consumed, we can make some special indicator on top of it.

I'm not familar with UI code, so I'll just give what I know and hope someone can take care of it:

* `KC3Ship.prototype.findDameCon` finds the first damecon of a ship.

```javascript
	   /* FIND DAMECON
	   Find first available damecon.
	   search order: extra slot -> 1st slot -> 2ns slot -> 3rd slot -> 4th slot
	   return: {pos: <pos>, code: <code>}
	   pos: 0-3 for normal slots, 4 for extra slot, -1 if not found.
	   code: 0 if not found, 1 for repair team, 2 for goddess
	   ----------------------------------------- */
````

* `KC3Fleet.prototype.getDameConCodes` does the same thing for a fleet. but only return an array of code (return array is always of length 6)

    example: `[2,1,1,-1,-1,-1]` means flagship can use a repair goddess when necessary. and 2nd, 3rd ship can use repair teams. last 3 `-1` means either the ship is not equipped with damecon or there is no ship in that place.

* new battle prediction result format:
    I think it's more straight forward to just give examples: https://gist.github.com/Javran/c8e36378e8a70e1959162843847ca865

dragonjet:
Close #1292


dragonjet:
Unfortunately unable to test this, and I'm sure non of our testers would want to spend damecon just for this that event is upcoming too. So I'm gonna take the risk of releasing this to public, and just hearing what they observe.

If I have time though, I will spend my last damecon, record the API and probably do some procedure like the one mentioned in #1292

dragonjet:
Deleting branch for cleanup.. For updates, create a new branch: `dev/battle-prediction-2`

PR #1298, opened 2016-04-17
Show Taiha HP left tooltip on panel

### LINKED ISSUE #1289: hp left to taiha(大破) (on panel)
suggested from DCinside by anon

sinsinpub:
Close #1289
Dislike to calculate 25% hp manually for each ship either, so add a tool tip to current HP of ship list panel :sweat_smile: 
Hint by #1289 , add a left value too and only appear when Taiha HP < Current HP < Max HP.

BTW remove the duplicated AACI tool tip builder codes, commit it here...


dragonjet:
hm there's something with the TL submodule

PR #1294, opened 2016-04-16
[dev-show-aaci] Show AACI battle flag on panel

sinsinpub:
Close #959

For #959 as the comment, I'm adding the AACI icon to the battle flags row of the battle activity panel like this:
image

And two dark theme icon just missing :joy: 

btw this fix the KO stamp for PvP to correct condition, the mistake Ive made on last PR. :sweat_smile:

PR #1308, opened 2016-04-22
[dev-show-aaci] Show AACI battle flag on panel #2

sinsinpub:
Continued for #1294 

Fix twice triggered AACI node issue.

And for @Javran and @dragonjet :
I've considered the way that show AACI indicator in ship box at first. But there are some design problems:
 * Battle activity attach the data for current battle node only
 * Shipbox list attach the data from player's fleets (showing current selected)
 * Extra data for indicators have to be passed to Shipbox object (or every ship objects in Fleet array) and kept during the battle node completed. Otherwise the indicators would disappear if user click on other fleet and switch back (or other events re-render the shipbox)
 * I add a extra attribute to ship object for the MVP icon, and do cleaning for each time battle node changed.
 * DJ gonna to pass extra data to Shipbox define function to show Damecon consuming. I'm not sure they will 'disappear' or not if user change current fleet during battle.
 * Finally I feel it's rather easier to show AACI (the battle node data) at battle activity panel, than pass it to Shipbox panel.

Maybe we have to get the proper way to pass the arguments to Shipbox for representing the 1-node only indicators on ship list panel.

### DIFF
--- a/src/library/objects/Node.js
+++ b/src/library/objects/Node.js
@@ -250,6 +250,7 @@ Used by SortieManager
 		this.eships = enemyships;
 		this.eformation = battleData.api_formation[1];
 		this.eParam = battleData.api_eParam;
+		this.eKyouka = battleData.api_eKyouka;
 		this.eSlot = battleData.api_eSlot;
 
 		this.supportFlag = (battleData.api_support_flag>0);
@@ -282,8 +283,8 @@ Used by SortieManager
 		
 		this.airbattle = KC3Meta.airbattle( planePhase.api_disp_seiku );
 		
-		if(!!attackPhase){
-			this.antiAirFire = attackPhase.api_air_fire;
+		if(!!attackPhase && !!attackPhase.api_air_fire){
+			this.antiAirFire = [ attackPhase.api_air_fire ];
 		}
 		
 		// Fighter phase 1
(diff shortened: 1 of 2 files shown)


dragonjet:
on #1297, I pass extra variable for damecon consumed, and it's only getting the .currentNode() so when fleet updates and it's not in the node that consumed the damecon anymore, it should disappear.

I also encountered that problem but opted not to attach extra data into the ship objects, because they might get included in the JSON encoding and saved into localStorage. One possible solution is to attach it as a function:
`Ship.something = function(){ return MyData; }`

however, jshint won't allow us if that Ship is inside a loop :laughing:

sinsinpub:
> it's not in the node that consumed the damecon anymore, it should disappear.

in this case it's true. But if user click on other fleet, and then switch back before the battle node finishing, is the data still there, should it disappear?

dragonjet:
i haven't actually tested it, coz that means ill need to use another damecon, and i only have 1 left in my alt account XD

but in theory it should not disappear before even if Fleet listener is called, `SortieManager.currentNode()` still has the damecon data. Only when it points to a new Node object it will disappear

sinsinpub:
So if that's true, invoking `SortieManager.currentNode()` from `Shipbox` to get the node data is better way. Maybe should change MVP icon data passing too :smiley:

dragonjet:
the problem is, Shipbox doesn't know what index it is rendering XD
so even if you know the damecon, and fleet info, you don't know which index to check :laughing: 
I've thought about passing the index to the Shipbox but I reverted back to just passing the data, I forgot why

PR #1312, opened 2016-04-23
[dev-quest-complete] Force complete quests on reward

### LINKED ISSUE #933: Quest stuck
This bug just happened here, using version 21 from webstore. The daily modernization quest got stuck. I already completed a bunch of quests today so it isn't like every single quest is doing this. I managed to capture these two logs on the console, not sure if they are related but my guess is they are:
sem titulo

EDIT: it happened to another quest just now, I think this whole 'kc3 session' is bugged right now and won't work until I reopen it

dragonjet:
Close #933


dragonjet:
anyone else seeing disabled merge button even though the status checks passed?

Issue #1279, opened 2016-04-12
[Bug Report] Timer ticks freeze randomly


mikenzb:
Seems like I might be the only one having this issue but I shall make a report anyways. It used to not bother me so much as it was more rare but it has been happening for 3 weeks and that is why there is a `debug-timers` branch however this issue has randomly started to occur a lot often now and it is driving me crazy.

Since DMM changed their API data last patch I was unable to use `debug-timers` branch to test but I added the 3f86ecae5b1b0ed2f29b1090e09dcfc68c73f827 commit to my local masters branch and I got it to work.

**Bug Report:
Date Seen:**
12-04-2016

**Version:**
KC3 Kai version: 27.5 Devbuild - Commit 9325af492530245ae61937cae0550f3cc04596ae
Browser: Google Chrome Version 50.0.2661.66 beta-m (64 bit)
OS: Windows 10

**Bug Description:**
Timer ticks freezing with no errors showing up
Please refer to the screenshots provided below.

**Severity:**
Moderate

**Steps I think to Reproduce:**
1. Start KC3 Kai and KanColle as usual.
2. Go into sortie
3. AFK in sortie for 10 minutes (?)
4. Come back and see timer ticks frozen (?) 
5. Expeditions timers no longer ticking and new timers aren't showing.

Note: does seem to happen randomly tho. 

**Workaround:**
Close KC3 Dev panel and reopen it and go to home port, issue resolved as the panel is reloaded, the game doesnt need to restart.

**Attachments:**
Console shows no signs of errors however timer ticks just stop
![timer1](https://cloud.githubusercontent.com/assets/5982780/14449710/91404928-00ca-11e6-9887-5c87fd8ed1c9.PNG)

Expedition timers out of sync:
![timer2](https://cloud.githubusercontent.com/assets/5982780/14449712/9eecc4e8-00ca-11e6-8f16-3b2871a044fe.PNG)

dragonjet:
that is very very odd. timers stopping without any warnings or errors.

mikenzb:
Issue still happens after #1325 not quite sure what is still causing it but don't think too hard about it because it seems like I am the only one affected by it.

Looked in the console but I don't see anything that stands out however it seems like some console messages aren't shown because I didn't have the console open...

dragonjet:
@mikenzb if satisfied with @sinsinpub's fix at #1328, you can close ;)

mikenzb:
[...]

PR #1325, opened 2016-04-26
Hotfix for unexpected timers stop issue #1279

sinsinpub:
Try to solve timers unexpected stopping related to #1279

Extra:
* Mute quotes looking up and transforming mass log
* Move quote looking up debug log to api.js
* Add new types 36 and 37 to gear tab of Strategy Room

### DIFF
--- a/src/library/modules/Meta.js
+++ b/src/library/modules/Meta.js
@@ -301,8 +301,6 @@ Provides access to data on built-in JSON files
 		
 		// Subtitle quotes
 		quote :function(identifier, voiceNum){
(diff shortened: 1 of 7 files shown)


dragonjet:
for `transformQuotes: function(quotes, checkKey)`, seems you're accepting a new param, but are the one's calling it required to adjust too, or is it optional param?

sinsinpub:
@dragonjet it's a option now. If there is a invoker would like to know about it, could set the option `true`.

dragonjet:
got it :+1:

PR #1326, opened 2016-04-26
[expcalc-goal-template] Leveling Goals Template

### LINKED ISSUE #1112: Adding goal template to exp calculator
I want to add a goal template to exp calculator:

A goal template can save Map & Rank & FS & MVP configurations, together with a ship type filter. User can have many goal templates. The idea is when user adds a ship to "Current Goal" section, we can pick up one of those templates that ship type matches. And apply its preset values to it.

For example: we can have following two templates (omitting some flags):

1. "BB CV" 3-2 S
2. "DD CL" 4-3 S

when adding a battleship, template 1 will be applied.
but when adding a destroyer, template 2 will be applied.

there are some tentative plans:

* highlight ships that one template can be applied to. for example, when selecting template 2, all DDs and CLs not listed in current goal gets highlighted. this could be a replacement of ship type filter (seems to be a planned feature).
* recommended templates, some common places are well known for leveling, when there's no template saved. user can choose to load these templates.
* apply a template to all ships applicable in Current Goal section (overwriting current settings)
* enabling / disabling / move up / down templates, since we will always use the first applicable template.
* might need some modifcation to "Current Goal" section, allowing user to apply one template to current leveling ship.

self-assigned for now.

dragonjet:
Close #1112

This is @Javran's work towards #1112

I'm just finishing it up with minor UI touches


dragonjet:
looks harder to style than I thought XD gonna move things around

Javran:
@dragonjet  hm, seems like the UI haven't been tuned, did you forget to push the changes?

dragonjet:
aye, but it's only color changes for now. havent even commited it

PR #1327, opened 2016-04-26
event history default to Spring 2016

Javran:
merge this during next maintenance xD


dragonjet:
we can merged it now XD
it's not Winter 2016 anymore either :laughing:

PR #1328, opened 2016-04-27
Refactor for unexpected timers freeze issue #1279

sinsinpub:
Another try to solve timers unexpected stopping related to #1279

As there's no way to restart, even inspect or check the existed browser timer object at all... the only way left is clear old one (if possible) and re-set a new one when timer stopped unexpectedly.
This time refactor the codes initializing and handling the Morale and UI updating timers. When go back to home port, automatically check if timer has not run for >2s, if true, reset it.
And this check is able to be triggered manually by clicking on morale timer on panel.

ps. applied to natsuiro theme only, not plain yet.

### DIFF
--- a/src/pages/devtools/themes/natsuiro/natsuiro.js
+++ b/src/pages/devtools/themes/natsuiro/natsuiro.js
@@ -3,8 +3,7 @@
 	_gaq.push(['_trackEvent', "Panel: Natsuiro Theme", 'clicked']);
 	
 	// Mathematical Constants
-	var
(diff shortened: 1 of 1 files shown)


mikenzb:
Yup tested out this branch and verified my issue has been fixed in a way, my timers still freeze but once I refresh to homeport the timers will come back and they won't freeze anymore. I'm happy with this hotfix for the time being.

dragonjet:
ah perfect, merging now for v30

Issue #1140, opened 2016-02-11
Ex-Item does not counted towards Equipment


ReiFan49:
## Problem Details
**Operating System**: Any
**KC3Kai Version**: Webstore/26.0
**Description**: item that placed on extra slot does not count towards ownership of the equipping ship

## Screenshots
In-game and Development Panel: (With **only** damegod that is equipped on ex-slot)
![image](https://cloud.githubusercontent.com/assets/9847780/12975297/18637b4a-d0ed-11e5-8fcd-631439c13120.png)

Before Equipped on Ex-Slot:
![image](https://cloud.githubusercontent.com/assets/9847780/12975330/557ec296-d0ed-11e5-8f5a-b304d4d07a7c.png)

After Equipped on Ex-Slot: (**Damecon** on basic slot, **Damegod** on extra slot)
![image](https://cloud.githubusercontent.com/assets/9847780/12975355/782cc568-d0ed-11e5-9913-019d49e98a5e.png)

sinsinpub:
May fixed by #1339

PR #1339, opened 2016-05-02
Update and fix Equipment List related

sinsinpub:
Fixes:
 * Scraped or consumed gear not updated to local storage in time

Updates:
 * Add support to check item from ship's ex-slot
 * Add type 38 for new fighter

### DIFF
--- a/src/library/modules/Kcsapi.js
+++ b/src/library/modules/Kcsapi.js
@@ -1378,6 +1378,7 @@ Previously known as "Reactor"
 				});
 				KC3GearManager.remove( itemId );
 			});
+			KC3GearManager.save();
 			KC3Database.Naverall({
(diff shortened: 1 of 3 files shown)


src/library/modules/Kcsapi.js
```diff
@@ -1377,6 +1377,7 @@ Previously known as "Reactor"
 					rsc[i] += x;
 				});
 				KC3GearManager.remove( itemId );
+				KC3GearManager.save();
```
dragonjet: we could probably move the `save()` outside of the loop

sinsinpub: right, amended

dragonjet:
:+1:

sinsinpub:
btw, the item on ex slot issue has been reported at #1140

Issue #1324, opened 2016-04-25
TP calculation


Javran:
we can do TP calculation when using transport escort combined fleet this might be useful for the coming event.

related documents:

- [Wikia: Events](http://kancolle.wikia.com/wiki/Events)
- [WhoCallsTheFleet TP calculator](http://fleet.diablohu.com/calctp/)

I don't have much time working on this before the event, so feel free to take over if you are interested.

ReiFan49:
how about wikiwiki one ?

Issue #1137, opened 2016-02-11
[Suggestion] Heartlock Reminder


ghost:
I think one potentially useful feature would be to a reminder players to heartlock rare ships or event reward ships after they return to base, if they don't have one already. I haven't done it myself but I've seen a lot of people scrap or modernize away their new drops or event reward ships after the map because they forget to lock them. The list can be pretty subjective, though.

ethrundr:
@KC3Kai/developers this might be a good idea, to prevent accidental modernization or scrap, especially for the reward ships since it'll be very hurtful to lost them in such way.

ReiFan49:
since `dev-api` is simple, after `dev-ledger` and `dev-api` is merged, i'll prepare the ship filter which includes the old feature that marks a ship as important/salty one if dropped will keep remind the player until heartlock api is called upon the respective ship...

ethrundr:
don't forget to let the user to choose manually which ships they want to be warned as well

ReiFan49:
![image](https://cloud.githubusercontent.com/assets/9847780/13028200/a5625da0-d29a-11e5-8973-a7b7b5595da8.png)
in-fleet reminder, still cannot think about ship that is not on fleet...

Javran:
what about activity tab?

ethrundr:
Why don't do this:

Brings out a pop-up for each time the user access port, notify that there is a ship that still unlocked.
this way we can notify when the ship is on fleet or not. Of course this only can be done if we can know exactly which ships is locked even if they're not in fleet.

Then, on the Strategy Room, add another tab to let the user picks ships that they willing to be notified. unlike the leveling tab, which shows only ships the user had, this tab will show ALL ships regardless what we have or not. the tab will be divided into three section:
- Available: which contain all unlocked ships and status whether we already have them or not
- Notify Me: which contain all "available" ships that has been selected by the user to be notified
- Locked: which contain all currently locked ships. this section is "read-only" (the user will be unable to modify the list).

When there is a locked ship being unlocked, then we'll create another pop-up notifying the user about the unlock. On the list, this ship will be moved to the Available section.

dragonjet:
or, since we already have an "update" detector, we can check if the non-heartlocked ship is in the latest updates. if yes then notify

I also agree with @ethrundr's idea, because you know people will be hunting graf or mizuho from the event, which they want to be notified for heartlocking even if it isn't a recent update

ReiFan49:
[...]

Issue #1062, opened 2015-11-22
Save fleet composition


Javran:
This is a feature request.

Can we have something in strategy room that saves current fleet compositions?
The idea is to allow user to keep fleet and equipment configurations, name and share them.
This could be extremely useful for events - you need to remember your E1, E2, etc. configs, for last dance or for farming ships etc.

With feature implemented in #754, it's also good for sharing preset configurations - not just wikiacode

dragonjet:
:+1:

Javran:
just bumping this thread. I think this might be useful for incoming events.

Javran:
assigning myself working on this. see branch [dev-fleet-viewer](https://github.com/KC3Kai/KC3Kai/tree/dev-fleet-viewer)

Javran:
already done long time ago, closing.

PR #1340, opened 2016-05-03
Release 30.0.1

dragonjet:
#### Already included
* [Updated translations](https://github.com/KC3Kai/kc3-translations/commits/master)
* Updated event map node letters (ongoing)
* Update and fix Equipment List related (PR #1339 by @sinsinpub)
  *  Scraped or consumed gear not updated to local storage in time
  * Add support to check item from ship's ex-slot
  * Add type 38 for new fighter

#### Goals to include
* #1062 Save fleet composition
* #1137 Heartlock reminder
* #1324 TP Calculation


dragonjet:
webstore update failed. the version number in `manifest.json` is still `30.0.0` which webstore did not accept. we should update it now to `30.0.2` to do the next release, ill do this quick

dragonjet:
sorry about that, the first thing I should do every release PR is to update version numbers

Issue #1223, opened 2016-03-25
[Suggestion] Adding new formula for LoS calculation


HKGPsy:
The new "33 Los formula" had been shown to be more accurate than the old formula. It would be nice if the next update can include this new formula as a reference for player.

Here is the source for the formula
http://ja.kancolle.wikia.com/wiki/%E3%83%9E%E3%83%83%E3%83%97%E7%B4%A2%E6%95%B5

c933103:
Sth similar is also available in Formula 33 section in EN Wikia's LoS
section. (despite indirect).

Note that other viewer like poi viewer have already included the new LoS
calculation method.
2016/03/25 14:46 "HKGPsy" <notifications@github.com>:

dragonjet:
Merged #1352

PR #1352, opened 2016-05-07
Formula 33 LoS calculation 

BossaGroove:
Did not set as default, please review and make adjustments if needed

https://github.com/KC3Kai/KC3Kai/issues/1223
https://github.com/KC3Kai/kc3-translations/pull/205


src/library/objects/Fleet.js
```diff
@@ -398,6 +398,7 @@ Contains summary information about a fleet and its 6 ships
 		switch(ConfigManager.elosFormula){
 			case 1: return this.eLos1();
 			case 2: return this.eLos2();
+			case 4: return this.eLos4();
```
BossaGroove: Did not set as default, please view and make adjustment if neededDid not set as default, please view and make adjustment if needed

src/library/managers/ConfigManager.js
```diff
@@ -186,7 +186,7 @@ Retreives when needed to apply on components
 		
 		// Toggle Equipment LoS
 		scrollElosMode :function(){
-			this.elosFormula = (this.elosFormula % 3) + 1;
+			this.elosFormula = (this.elosFormula % 4) + 1;
```
BossaGroove: I think no need to hard code the number here, but changed anyway

dragonjet:
Thank you for this :+1: :+1:

Issue #1356, opened 2016-05-09
localStorage limit exceeded


Javran:
I think this is critical problem that has to be addressed quickly:

![2016-05-09_034802_3840x1080_scrot](https://cloud.githubusercontent.com/assets/1096354/15106695/16d9d8b4-1599-11e6-91be-82ead9de9cee.jpg)

somethng has occupied all space of localStorage, making KC3 itself unusable (devtools cannot be initialized properly because of this)

the sources of this might be:
* `localStorage.raw`: by removing `localStorage.raw` I managed to enter the game with devtools initialized properly, but the error message will appear whenever entering main screen or do something related to ship list.
* `localStorage.ships`: might not be a big issue for now, but I feel something is wrong with it:
![2016-05-09_035721_3840x1080_scrot](https://cloud.githubusercontent.com/assets/1096354/15106894/45a2a080-159a-11e6-8c08-961024db8049.jpg)

not sure what `lastSortie` is used for, but it's definitely taking many spaces. and the variable name `lastSortie` suggests this array might be intended to keep only few of last sorties instead of accumulating them all...

ReiFan49:
`lastSortie` should only retain it's latest sortie and sortie0 as placeholder, maybe there's some design flaw (i should recheck this),
also `chrome.storage.local` is the other solution for non string based item store (but it means migrating and yeah, whole change on the way storage system works)

Javran:
update: might not be that urgent after all, I have `localStorage.master`, where some parts of `api_start2` is stored in the past. clearing localStorage works for me.

but I think this problem still need to be solved, as chrome has hard-coded localStorage size quota to 5MB, and we have at least used half of it:

![2016-05-09_050808_3840x1080_scrot](https://cloud.githubusercontent.com/assets/1096354/15108464/11b11900-15a4-11e6-9bae-b04b2801e3a5.jpg)

(see: http://stackoverflow.com/a/17887889/315302 , put code in your strategy room's console to see the actual size.)

here are few things I think we can do:

* move master data to IndexedDb or somewhere else
* remove `localStorage.raw` if detected, it now has no used and user might have the risk of using up localStorage space just as I did)
* `lastSortie` issue, as I'm writing a reply @ReiFan49 is aware of it and hopefully this one can be resolved soon.

sinsinpub:
> remove localStorage.raw if detected,
it's a typo. `.master` is the deprecated one... 

[...]

PR #1358, opened 2016-05-09
Local Storage Sortie Data Overquota

ReiFan49:
this is only workaround to force every `sortie0` data to not appear anywhere around the sortie0 itself,
**you need to supply/repair the respective ship to remove the ghost entries**

- Attempt #1356

to test the data there are some steps that need to check:

1. find any ship with long entry of `lastSortie` data<br>(it's not a must for a global maximum length one, it can be more than one)
2. output the current state of `lastSortie` of the respective ship
3. perform supply and/or repair
4. output the last state of `lastSortie` of the respective ship

### DIFF
--- a/src/library/managers/ShipManager.js
+++ b/src/library/managers/ShipManager.js
@@ -71,6 +71,22 @@ Saves and loads list to and from localStorage
 						cShip[key] = tempData[key];
 				});
 				
+				// enforce freshify sortie0 placeholder
+				(function(){
+					var szs = 'sortie0';
+					var ls  = cShip.lastSortie;
+					var cnt = 0;
+					var szi;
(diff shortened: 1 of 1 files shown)


dragonjet:
as far as I can understand, can't we just `cShip.lastSortie = { sortie0:"" };` on this?

ReiFan49:
no, it doesn't work that way, because it changes the whole structure for `lastSortie` itself.
it works to prevent non `sortieXXX` keys on `pendingConsumption` being treated as the last one or something, `KcsApi.js` that takes the use of `lastSortie` itself, so it can prevent empty `pendingConsumption` but damaged ship being repaired and resulting an error.

dragonjet:
merging this before this happens to our users

PR #1368, opened 2016-05-11
[dev-cleanup] Remove unused pages and code

dragonjet:
* Removed wikia export from strategy room
* Removed presets (not really useful and just takes up `localStorage` space). Fleet manager does this better now.
  * Removed recording into `localStorage`
  * Removed tab from strategy room
* Removed `pages/help`
* Removed Strategy `exops`
* Removed Strategy `questlist`


Javran:
maybe we could put wiki code exporting to fleet manager?

dragonjet:
hm tbh I wanna remove wikia-specific features. not everyone uses wikia and not alot are using the template, and iirc the wikia itself made a new template for reporting.

sinsinpub:
I think we could keep the preset API codes as a placeholder or API document as long as official API not change it. just skip the proceeding code such as local storing behavior.

And I have a plan to build a "Quest Library" page, `questlist` might be the same purpose created before. :laughing:

dragonjet:
ah yes you may recreate them in future branches just in case. for these few versions while they're empty yet, we can remove them. there's a "copy-pastable" `_tpl` folder in strategy just in case

PR #1370, opened 2016-05-12
Toggle fighter power mode by clicking on panel

### LINKED ISSUE #1369: Fighter power calculation without airpro?
Because fighter power calculation for aerial support neglect airpro, is it possible to obtain the value from within kc3kai?

sinsinpub:
Close #1369

Formula mode config and calculation logic is already there, just add click event to two themes.

ps. `air_formula` is not added to `settings.json`, so it's not able to be configured at settings page yet. should add it?

### DIFF
--- a/src/library/managers/ConfigManager.js
+++ b/src/library/managers/ConfigManager.js
@@ -196,6 +196,12 @@ Retreives when needed to apply on components
 			this.save();
 		},
 		
+		// Toggle Fighter Power
+		scrollFighterPowerMode :function(){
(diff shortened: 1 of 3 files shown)


dragonjet:
how was `$(".summary-airfp").on("click",function(){` removed before. I know we have an on-click for that before to toggle both LoS and FP, I'm 100% sure. was it removed somehow at some point.

dragonjet:
also, we don't need it in settings page, just it being clickable is fine

PR #1382, opened 2016-05-18
Accurate landing craft bonus

Javran:
accurate landing craft income bonus, according to formula found on wikia:
2016-05-18_132741_3840x1080_scrot

now T89 & T2 and improvement bonus are all taken into account.

### DIFF
--- a/src/library/objects/Fleet.js
+++ b/src/library/objects/Fleet.js
@@ -241,7 +241,7 @@ Contains summary information about a fleet and its 6 ships
 				return;
 			improvedEquipmentCount += 1;
 			improveCount += stars;
-		};
+		}
 
 		var normalCount = 0;
 		var t89Count = 0;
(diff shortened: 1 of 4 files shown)


sinsinpub:
trying it, mine:
image

Javran:
@sinsinpub not seeing any problem, could you give more details?

sinsinpub:
@Javran only the negative ammo is a problem...

Javran:
@sinsinpub Exped38 doesn't give you ammo in the first place. the "-104" part is estimated resupply cost.

sinsinpub:
@Javran oh, I misunderstood it... tooltip why not show up gain part result to match with the SR history page, not only the formula? eg: `480(=400*1.2) - 104`

Javran:
@sinsinpub I'm not sure about expedition history, that's a separated function. maybe the numbers in expedition history come just from KanColle API, which doesn't take re-supply consumption into account.

dragonjet:
seems all good? true the history only shows the actual result, and without resupply calculations, and acts like a predictor of some sort as well. gonna merge this :+1:

sinsinpub:
to clarify the net income value, just change the tooltip a little bit...?
image

Javran:
@sinsinpub I'm good with this change. :+1:

PR #1384, opened 2016-05-19
[dev-sroom-update] Improve update mechanism of Strategy Room tab

### LINKED ISSUE #1381: Ship Library problems.
in the lastest master branch: https://github.com/KC3Kai/KC3Kai/commit/6128991473c4ff0c2cf709ee4b30a1d4ae4285a1

wrong object layout:

2016-05-18_105110_3840x1080_scrot

also I find in this commit click on a ship also moves that entity to the top of the list, is it possible to disable this?

sinsinpub:
Close #1381 

The re-executed rendering and hash parameters handling problems has been realized when I tried to improve ship list page. It's not only the thingy about 'Ship Library', but an important mechanism all tab pages should rely on.

This time add a new term `update` to tab definition first, allowing each tab itself to decide if partially updating content required and how updating performed.

Currently `mstship`, `mstgear`, `quotes` supported,
More possible tabs: `ships`, `gears`, `fleet`, `aircraft`, `maps/event`

ps. performance of quotes tab is better by moving en/jp quotes loading into `reload` term.

### DIFF
--- a/src/library/objects/StrategyTab.js
+++ b/src/library/objects/StrategyTab.js
@@ -56,7 +56,8 @@
 	 * Terms of callback function definitions:
 	 *   "init": Invoked for the first time tab initialzing, only once (for once browser refresh).
 	 *   "reload": Invoked when data loading required, optional once, may many times on demand.
-	 *   "execute": Invoked for rendering HTML content for each time tab is shown.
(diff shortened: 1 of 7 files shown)


sinsinpub:
Two float buttons added to right bottom for convenience:
image

Javran:
tested on my side, good work :+1:

PR #1385, opened 2016-05-21
Show operational radius of land base aircrafts for Equip List

sinsinpub:
* Gear master property `api_distance` used, new `or.png` stats icon added
* Fix a sorter issue

Official master API data is ridiculous... values of `api_cost` and `api_distance` re-use the ones of previous item instead of reset to 0, even it's not a aircraft. That means from the one which `api_id` is geater than 15, all left items except aircrafts are holding wrong `api_distance` :joy: 
So we have to limit the `api_distance` only applied to those types which land base is able to use (currently `[6,7,8,9,10,37,38]`)

And new `or.png` icon reuses the `ac.png`, could replace it if any other better one available.


sinsinpub:
Screenshot (ordered by operational radius):
image

dragonjet:
Deleting branch for cleanup.. For updates, create a new branch: `dev/plane-distance-2`

PR #1397, opened 2016-06-01
Support api_req_kaisou/slot_deprive

dragonjet:
support the new mechanic to swap equipment across ships


sinsinpub:
it seems that not a 'swap'...? I changed an equip which hold by another ship, the equip hold by source ship did not go to the 'another ship', just become free equippable  :sweat:

dragonjet:
lol the naming `deprive` is funnier from the devs, but yea, we could change comments on the files

PR #1421, opened 2016-07-06
view level-dependent ship stats

Javran:
Added a slider in ship library, this allows checking level-dependent ship stats at specific levels, namely ASW, LoS and evasion:

2016-07-07_061159_3840x1080_scrot


dragonjet:
wow nice :+1:

sinsinpub:
suggest using icons instead of 'asw, los, eva'?

Javran:
@sinsinpub that'll be better but I don't have a good idea about how to place them.

sinsinpub:
how about?
image

Javran:
@sinsinpub sorry for late response, good work :+1: could you create a PR for this?

Issue #1365, opened 2016-05-11
[Suggestion] Disable annoying taiha alert on top screen when screenshotting


tayuku:
Hiya! I don't know if this feature is fully supported or important. 

Is there a way to disable the top portion of the annoying taiha alert?
Maybe there's another option to only have the sound effect and bottom part flashing?

Thanks a lot!

dragonjet:
ah there's no feature like that yet, but we can add it

rephira:
@dragonjet I suggest that we change this further and make an option where the taiha alert shows up only on the dev tool and not the game screen.

Some people use KC3kai and like the fact that there is an alert, but says that when they are doing live streaming, it can get in the way.

DLRevan:
Can I suggest a few more things if the aim is to improve this function in the future?

There can be options to mute and turn off the overlays in the dev tool area, just like there is a mute sound button. To reduce the chance of user forgetting its off, the dev tool could still flash.

Also, right now the alert triggers when going into night battle if ship became taiha in day during on the same node because hp updates again when player clicks night battle. I think it would be better if a check is made so that it only triggers after battle result.

dragonjet:
Fixed in #1424

for @DLRevan's additional suggestions, please create a new issue.
thanks :+1:

Issue #1059, opened 2015-11-21
Mute taiha alert heartbeat


AlanSchaeffer:
Oh god can we have a button to temporarily mute the current taiha alert? I just had to watch the entire E-3 aftermath, plus two expeditions arriving and then about 6 ships worth of repairs to finally turn the alarm off, I think I may be going insane now :stuck_out_tongue:

mikenzb:
There's currently a temporally way to disable the annoying taiha alert when your in a sortie, just pick a fleet that isn't in the sortie (support expedition/expeditions fleets), and it will disable it.

AlanSchaeffer:
Why didn't I think of that XD

ethrundr:
All in all, having mute feature to temporarily mute the alert still a good option.

dragonjet:
:+1:

dragonjet:
Fixed in #1424

Issue #1005, opened 2015-11-04
Screenshot without taiha alert blur


bcc32:
Currently, taiha alert blur seems to mess with screenshots.  Is there a way to make screenshots temporarily disable blur, or to take them _through_ blur?

dragonjet:
hm, that's a problem. one thing is that we can't take it through blur - but possibly the one-tick off, but may cause more problems to low-end comps XD

dragonjet:
Fixed in #1424

Issue #980, opened 2015-10-28
Add Taiha Alert settings for low-spec computers


NicolaosGreene:
If I have Annoying Taiha Alert on and a ship gets Taiha'd I cannot see anything but my background color. I can still click on things but I can't see anything.

ethrundr:
That's why it is called "Annoying" :laughing:

LostLogia4:
Can we get a screenshot of it?

dragonjet:
you could use the devtools inspector to select `class="box-wrap"`
then on the styles list, see under `.critical`
disable one of each css style and see which line will bring back the visuals

NicolaosGreene:
disabling -webkit-filter: blur(1px); returns the visuals. Furthermore, when I played on my desktop (which is significantly more powerful) I also noticed there was a sound effect that was to go with it that seems to be dead on my laptop, I suspect that this feature may be not usable on low end machines. Because of this you may want to have an alternate version that makes the side panel red like it does now and possibly sends an alert for those who run on a low end devices.

ethrundr:
The best solution is to let the user modify what they want to stay active as the alert (sounds, flashing red on panel, bloods, blur) after they activated the alert, and one of those feature should be selected, so they will still have the Taiha alert, but on their own preferences.

This is to compensate those who are playing on low-end devices, or while doing resource-intensive works such as scanning the computer for viruses or copying huge chunks of data in or out of the device internal harddisk.

Alternatively, instead letting them choose, we make another preset (let say only include the flashing panel and sound), called as "Annoying Taiha Alert - Low-spec Mode", or similar to it, that is predefined (user cannot change it by themselves without altering the code) by the code. This submenu should only be appear after the main "Annoying Taiha Alert" has been selected and activated. When the "Annoying Taiha Alert" is deactivated, then this submenu will be gone altogether until they activate the menu again.

dragonjet:
yea additional settings should suffice, though I was originally planning to leave it be coz I thought I was looking at a relatively small group of people using that - as it is by default disabled.

ethrundr:
@dragonjet if you look to my group, lot of them are expecting that bloody alerts, dubbing it for their secretary "got her period" :laughing: 
and yeah, most of them has low-spec laptops (not even a computer, but a laptop instead, which usually far lower than a computer)

ethrundr:
reminding about this one.
more and more people complained about the memory usage of current alert, so it's about time I think we finish this one

@dragonjet

[...]

PR #1424, opened 2016-07-12
Taiha Alert Settings

dragonjet:
would be nice if anyone can test in their current playthough. thanks :+1: 

aims to fix:
* #1365
* #1059
* #1005
* #980

ss 2016-07-12 at 10 29 52


sinsinpub:
A question: does it intend to reserve term `SettingsDisplayOnPvP` not used...?

dragonjet:
I planned on adding that setting, but it seems the issue about Taiha alert triggering during PvP goes deeper than just adding it to the if statement, so that should be on hold for now

PR #1436, opened 2016-07-18
Tracking quests with shared counter

sinsinpub:
Try to handle quests which sharing single counter, esp those quests are able to be activated at same time. (which cause on 1 condition full-filled, progress +2 for all shared quests)

Currently just added Bd6 and Bd5 (`[218, 212]`) pair, more configs are expected.
iirc some 'submarine sunk' quests also sharing counter.

### DIFF
--- a/src/library/managers/QuestManager.js
+++ b/src/library/managers/QuestManager.js
@@ -17,6 +17,8 @@ Uses KC3Quest objects to play around with
 		timeToResetWeeklyQuests: -1,
 		timeToResetMonthlyQuests: -1,
 		
+		sharedCounterQuests: [ [218, 212] ],
+		
 		// Internal constants for time period quests
 		_dailyIds: [201, 216, 210, 211, 218, 212, 226, 230, 303, 304, 402, 403, 503, 504, 605, 606, 607, 608, 609, 619, 702],
 		_weeklyIds: [214, 220, 213, 221, 228, 229, 241, 242, 243, 261, 302, 404, 410, 411, 613, 703],
--- a/src/library/objects/Quest.js
+++ b/src/library/objects/Quest.js
@@ -100,17 +100,28 @@ Quest Type:
 	/* INCREMENT
 	Add one to tracking progress
 	------------------------------------------*/
-	KC3Quest.prototype.increment = function(reqNum, amount){
-		if(this.tracking && this.status==2){    //2 = On progress
+	KC3Quest.prototype.increment = function(reqNum, amount, noLoop){
+		var self = this;
+		if(this.tracking && (this.status==2 || !!noLoop)){    //2 = On progress
 			if(typeof reqNum == "undefined"){ reqNum=0; }
 			if(typeof amount == "undefined"){ amount=1; }
-			if (this.tracking[reqNum][0] + amount <= this.tracking[reqNum][1]) {
+			var maxValue = (!!noLoop && this.status!=2) ? this.tracking[reqNum][1] - 1 : this.tracking[reqNum][1];
+			if (this.tracking[reqNum][0] + amount <= maxValue) {
 				this.tracking[reqNum][0] += amount;
 			}
 			KC3QuestManager.save();
 		}
+		if(!noLoop && Array.isArray(KC3QuestManager.sharedCounterQuests)){
+			KC3QuestManager.sharedCounterQuests.forEach(function(idList){
+				for(var idx in idList){
+					if(self.id !== idList[idx]){
+						KC3QuestManager.get(idList[idx]).increment(undefined, undefined, true);
+					}
+				}
+			});
+		}
 	};
-
+	
 	/* ISCOMPLETE
 	Return true iff all of the counters are complete
 	------------------------------------------*/


dragonjet:
hmm, but aren't we doing this in some quests? e.g. sink CVL. iirc what we should do on those cases are just trigger increment twice, one for each quest that needs it

sinsinpub:
hmm, but ive searched all js, and no twice or repeating quest `increment()` invoked found... :sweat_smile:

dragonjet:
`src/library/objects/Node.js`
at this time of writing is at line number 798 onwards :sweat_smile: 
multiple calls to increment for different quest IDs
```
KC3QuestManager.get(217).increment();
KC3QuestManager.get(211).increment();
KC3QuestManager.get(220).increment();
```

sinsinpub:
yea, i see these. They are just increase counter 1 quest (id) by quest. In fact for shared counter situation, each quest counter will be inc multiple times (for Bd6 and Bd5, sink 1 AP, counter of both quests will be added 2). These codes not support this yet.

And known fact: if one of shared counter quest is open but not activated, its counter will stop at target - 1. For Bd6 and Bd5, if Bd6 activated and Bd5 not, after sinking 5 AP, Bd6 is completed, and Bd5 is 90% (need 1 more AP).
ofc if both are activated, just sinking 3 AP will make them completed (just sink 2, Bd5 will be completed, and Bd6's counter is 4/5).

dragonjet:
ohh I see what you mean, yes I remember these cases where it still counts internally in-game as long as a similar quest is active. can more @KC3Kai/testers take a look?

nonsensed:
According to the english wikia, the next date bd6 is available is on the 22nd in Japan.

This is a small change I've been looking forward to, thanks ssp.

sinsinpub:
and according wikia, Bd8 and Bw5 may be the another 'buggy' quests pair...
if there were more counter shared quests group, let me update the config.

nonsensed:
What I tested tonight:

1. Activate Bd6 (0/5 transports) and Bd5 (0/3 transports)
2. Sunk 2 transports
3. Game counted 4/5 kills for Bd6, 3/3 for Bd5.

This PR works properly for that.

1. Did not turn in Bd5. (3/3 complete)
2. Turned off Bd4 (4/5 80% in-game)
3. Sunk 1 transport.
4. In-game, Bd4 is 80%, so 4/5.
5. Kc3 thinks it is 5/5. (incorrect)

There is also 1 more thing to test, but since my Bd5 is 3/3 already, I can't test it right now. (same as the 2nd example, except with Bd5 at 0/3)

I'm not sure about the submarine quests Bd8 and Bw5. I thought it was normal +1 even with both turned on. I'll test it when the weekly quests are reset.

sinsinpub:
what do u mean about Bd4...? :open_mouth: it's a 3 CV(L) sinking quest...

nonsensed:
My fault! Bd6.

sinsinpub:
image
i'm at the same situation now...let me try to debug it

sinsinpub:
hmm... my case: after sinking 1 more and then reopoen Bd6, it's still 4/5
image

nonsensed:
Ok, I'll try it again on the next time. I think it's on the 28th.

Actually, one possible difference in our tests:
With Bd5 at 3/3, viewing it in the quest list caused it to be removed from the kc3 panel. When I sank 1 more transport, Bd5 and Bd6 were both inactive in the panel, but Bd5 was active (3/3) in the game.

sinsinpub:
so i continued testing after Bd5 removed from panel, result still:
image

nonsensed:
@sinsinpub Where does the info about Bd8/Bw5 come from? I'm trying to find a way to make them +2 in-game, but so far I can only get +1.

There was some unexpected behavior from Bd8 after finishing Bw5/Bd8 in the kc3 client. I'll try to get more information before describing it, though. I think it only matters if the issue also affects Bd5/Bd6.

nonsensed:
Well, I broke it.
[video of test](https://youtu.be/wuXUL4dsnZ0)
Links to skip to important parts in the video description.

The only other thing I can add is: I reset all quests yesterday.

nonsensed:
C10 causes C3 to get +2 if both are active. If only one of them is active, they will +1 each other. Even if C3 gets +2, I'm not sure if C10 also gets +2.

Did not check if the same is true for C2.
(all of this is in-game behavior, not kc3)

sinsinpub:
@nonsensed yep, these are known behaviors mentioned at wiki already, many one-time PvP quests are sharing counter with C3. And C2 is always independent. We may add these C3 related data, but would not frequently used :sweat:

nonsensed:
> According wiki, C3, C5, C6, C7 may share 1 counter. **More verifying required** XD

Also, it was a new quest.

Is this the same wiki that told you the submarine quests shared a counter and worked the same way as the transport quests?

sinsinpub:
@nonsensed I'm always reading [wikiwiki](http://wikiwiki.jp/kancolle/?%C7%A4%CC%B3), and submarine one is also my 'iirc'... I guess I should cancel that as I've not been able to confirm it neither. :sweat_smile:

nonsensed:
Ah, thanks for posting the source. I appreciate it.

sinsinpub:
keep such a potential feature, only handle Bd5 and Bd6 for now.

Javran:
why is this PR suddenly be merged? this is nowhere near satisfactory: no confirmation of problems being fixed (at least I don't see in this thread), no a single line of comment is added in code to explain what's happening, and still I don't know why we should have this feature while it's a game-side bug and kc3 should have corrected its own counter by tracking quest progress (percentage)

sinsinpub:
sorry for late as my pad out of power...

 * no further problem is being fixed, so take it easy to merge such a patch. will not affect other quests at all.
 * there are also game-side bugs we already fixed on kc3-side. recently for example: the `三隅`

u can keep/remove the config array inside `QuestManager.sharedCounterQuests` to compare the difference of quest counter, because today we should have both quest Bd5 and Bd6.

Javran:
@sinsinpub would you mind writing down some comments in code, at least point others to a reference (I think wikiwiki or so would be fine).

I admit the `三隅` is a patch made from our side, but it doesn't have much impact anyway and it make things easier for translators. but share counter affects game logic, in some very unexpected and unpredictable way, to name a few:

- occasionally there are new PvP quests and they might share the same counter as daily quests, how do we deal with them?
- some users would also do quests on phone or have multiple clients, does this fix overshoot? for example, I sunk one/two transport on phone and later come back on pc to complete whatever remained, would quest counter bump twice because kc3 would never thought that we had made progress on some other clients?

also I don't think we have mechanism that turns a completed quest into incompleted one if it finds out that the progress it has tracked is ahead of the game (e.g. quest is `5/5` but it shows 80% in game), with the possibility of bumping a counter twice, this could happen. and this brings up another reason why I prefer not to fix it: if we only bump a quest counter once at a time, we can never overshoot, so our quest percentage check can always bring us back on track, but with this change, it becomes possible and if it happens user can do nothing but manually reset it.

sinsinpub:
this patch does neither impact much, only confirmed Bd5,7 affected. not deal with PvP quests if not able to confirm.
it just solved: the quests always in game complete/ahead but kc3 counter still late.
no relation with syncing of different clients, that problem always remains.

Javran:
let me restate the important part: if kc3 falls behind, it's okay as it can already resolve itself automatically. but try to bump a counter twice we risk to be ahead of the game, in which case we don't have a fallback mechanism.

I know this only affects Bd5 and Bd7, but don't you think this brings in more complication than the problem it tries to solve?

Javran:
well I'm just complaining, this has been merged and efforts spent on this would better not be wasted.
if you can just put some comments in code, include some references and maybe explain what the code does (especially what is that `noLoop` thing), and I think I can live with this.

sinsinpub:
yep, remark will be added later.
it's not just simply bumping counter twice, it's with conditions...  for those 2 quests, be ahead of game is not expected. if found, just think it a bug i should fix.
and this kind of game-side counter bug is indeed complex for client-side not easy to solve lol. so it failed...

PR #1441, opened 2016-07-20
Highlight ships that are close to remodel or can be remodelled

DynamicSTOP:
**Problem** young teitoku got too many ships inside recommendations section. Can easily miss ship that can be remodelled or close to remodel.

**Solution**
- highlight with green colour ships that can be upgraded
- highlight with blue colour ships that are close to upgrade (lvl difference could be changed in settings)

[How it looks like in the end]( http://imgur.com/00gVjUI)

P.S. `"min" : 0` actually isn't working.

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -664,5 +664,24 @@
 				"options" : { "label" : "%" }
 			}
 		]
+	},
+	{
+		"section":"Strategy Room",
+		"help":"",
+		"contents":[
+			{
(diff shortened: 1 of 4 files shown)


dragonjet:
ah perfect, thanks for this :+1: 
please also reply here up to 5 links that we can put under your name for the credits section

DynamicSTOP:
https://github.com/DynamicSTOP
https://www.facebook.com/DynamicSTOP

Thanks!

PR #1444, opened 2016-07-21
Update akashi.json

nonsensed:
new akashi upgrades from the July 15 patch


dragonjet:
by the way, I also added you now to the dev team which has write access to the main repo in case you need updates like these

PR #1456, opened 2016-07-28
take into account improvment stat for fighter-bombers

### LINKED ISSUE #1455: [Suggestion] Effect of Akashi's improvement on AA of Fighter and Fighter-Bomber plane
For each level of improvement (star), there would be 0.2 additional AA value for carrier based fighter plane and 0.25 for fighter-bomber plane

It would be nice if the fighter power calculator can take this into account for more accurate calculation

Source: 
Carrier based Fighter plane: [http://ja.kancolle.wikia.com/wiki/%E3%82%B9%E3%83%AC%E3%83%83%E3%83%89:951#26]
Fighter-bomber plane: 
[(http://ja.kancolle.wikia.com/wiki/%E3%82%B9%E3%83%AC%E3%83%83%E3%83%89:951#32)]

Javran:
close #1455

### DIFF
--- a/src/library/objects/Gear.js
+++ b/src/library/objects/Gear.js
@@ -51,15 +51,30 @@ KC3改 Equipment Object
 		// Equipment did not return on plane check, no fighter power
 		return 0;
 	};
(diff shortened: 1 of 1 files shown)


dragonjet:
:+1:

PR #1470, opened 2016-08-08
Leveling tab improvements

### LINKED ISSUE #1445: regarding highlighting close-to-remodel ships
I feel there are several things can be improved:

- leaving this strategy page on seems to cause some moderate CPU usage, I'll recommend leaving an option for user to decide whether they want it to be blinking

- there are several ships whose level is beyond remodel-level, (or has a cyclic final form) and these ships keep blinking regardless of the setting, I guess this is not an intended behavior:

2016-07-22_005200_3840x1080_scrot

- I've tried changing leveing difference to `-1` and `0`, but in both case ships mention in the bullet above are still blinking.

Javran:
few improvements on leveling tab, also close #1445 

- remove blinking animation
- let RemodelDb help with goal level recommendation

    this makes it possible to do things like recommending Lv 78 as a goal for Souryuu on Lv 31 
    (you could have remodelled  Souryu to Souryuu Kai at Lv 30,
    but in current master branch when the level goes beyond 30 there's no more recommendation)

- suggest updating current goal instead of removing a completed goal

    when user is removing goals for ships that have not yet reached their final form,
    pop up a dialog to ask user if they want to update goal to next recommended level
    instead of removing the goal.
    (example: removing Souryuu on Lv 31 when the goal level was Lv 30, ask user if they'd like
    to change the goal to Lv 78)

- highlight color improvement (make color of close-to-remodel and can-be-remodelled different)

- controls on top for showing/hiding sections and turning different kinds of highlight on/off

- ~~provide a way to hide Lv 99 goals~~
    (I think hiding "Other Ships" section is in some sense an alternative solution already)

- make highlight mutually exclusive
    at most one kind of highlight at a time, as there are ships that satisfy more than
    one highlight condition, we'll have conflicts if all kinds of highlights are shown.

- close-to-model level difference setting is now moved to expcalc tab
    the setting is only visible when "Close To Remodel" is clicked

- give a better look to "New template" button

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -664,24 +664,5 @@
 				"options" : { "label" : "%" }
 			}
 		]
-	},
-	{
-		"section":"SettingsStrategyRoom",
-		"help":"",
-		"contents":[
-			{
-				"id" : "sr_lvl_difference",
-				"name" : "SettingsLevellingDifference",
-				"type" : "small_text",
-				"bound" : {
-					"min" :  0,
-					"max" : 15,
-					"type": "Number"
-				},
-				"options" : {
-					"label" : "Lv"
-				}
-			}
-		]
 	}
 ]
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -670,18 +670,6 @@
 		"help":"",
 		"contents":[
 			{
-				"id" : "sr_lvl_toggle",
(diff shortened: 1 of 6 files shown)


Javran:
all done, feel free to merge.

dragonjet:
thanks :+1:

PR #1490, opened 2016-08-20
[dev-lbas-ui] LBAS i18n support

sinsinpub:
Done:
 * Replace messages with i18n terms
 * Fix name changing decoding
 * Fix stopped Exped faces refreshing


dragonjet:
ah perfect~ those html tooltips were added last minute and I type it knowing someone would ask it to be translatable but I didn't bother to i18n anyways coz scheduled release :sweat_smile:

PR #1492, opened 2016-08-20
Configurable Air Proficiency Values

dragonjet:
* Configurable Air Proficiency Values
* New settings types: `json`, `textarea`

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -28,19 +28,6 @@
 				}
 			},
 			{
-				"id" : "elosFormula",
-				"name" : "SettingsELosFormula",
-				"type" : "radio",
-				"options" : {
-					"choices" : [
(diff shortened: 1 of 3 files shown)


Javran:
@dragonjet what does "Min+Interception" mean? it doesn't seems to be used anywhere in code.

dragonjet:
ah it's the interception power implemented as part of LBAS interface.

calculation is at `Gear.js` `interceptionPower()`

it first gets interception power, takes adds the minimum value from the existing _min-max proficiency_ calculation

Javran:
@dragonjet I see, so interceptors have their own way of calculating air power?
now I feel it should be separated from other air formula options, at least it doesn't sound like alternative to whole number or min-max..

PR #1513, opened 2016-08-29
Prevent a call stack overflow when Math.stdev is called with too many arguments

Javran:
seems the Ledger is relying on `Math.stdev` for some calculation and I got an call stack overflow because of a long sortie history (~ 60000 elements). fixed by calling `Math.stdev([bool,] <an array>)` instead.

sorry for the noise in the diff view, the end of line gets messed up so I reindented them before working on this patch, the real changes are just 318796d and 39cd6ba .

I think @ReiFan49 is the original author of this part, could you take a look and see if this patch is good enough?

### DIFF
--- a/src/assets/js/global.js
+++ b/src/assets/js/global.js
@@ -655,15 +655,26 @@ String.prototype.hashCode = function() {
  - Population based deviation
 -------------------------------*/
 Math.stdev  = function(p1f /*, data*/){
-	var args = [].map.call(arguments,function(val){
+	// obtain a real array for carrying data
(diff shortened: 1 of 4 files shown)


dragonjet:
for me it looks good :+1: 

by the way, to view the diff without the whitespace, append [`?w=0`](https://github.com/KC3Kai/KC3Kai/pull/1513/files?w=0) to the end of the url

Javran:
let's been a while, I guess we are good to proceed and merge it.

dragonjet:
got it :+1:

PR #1520, opened 2016-09-02
Unify line endings of source files

sinsinpub:
No logic changed, just convert some files with mixed line endings to Windows/DOS format (CRLF) as most of rest files are.

For some historical reasons, some html/css/js files are edited by different developers under different OS or editors. The line endings are mixed with different styles (mainly LF and CRLF). I deliberately ignored it in the past to respect the old works and old contributors.
But personally I agree with the idea that @Javran mentioned to keeping identical text file styles of sources, such as line endings, indent white-space (tabs vs spaces), UTF-8 charset encoding, removing trailing spaces..

It's ok for me to use any style of these but the point is keeping identical. So this PR is to suggest that use identical line endings for this repo not only for single file, current is partial files in Unix (LF only) and partial files in Windows (CR +LF). Currently this PR just eliminates the mixed line endings inside a single file, other files are not changed yet.
And for @KC3Kai/developers, if no matter OS is Windows or Linux or OS X, remember to enable git config: `core.autocrlf = true`, git will help to convert your workspace automatically. See github official help: https://help.github.com/articles/dealing-with-line-endings/

For other styles, another discussion may be opened :smiley:


dragonjet:
but I have my tortoisegit autocomvert from crlf to lf :sweat_smile: let me see if I can change my settings, but I agree with the standardization :+1:

sinsinpub:
@dragonjet in fact, git do 'crlf to lf' when commit if `autocrlf=true`, git born for Linux, it's internal EOL prefers Unix LF. commonly `autocrlf=true` is suggested for Windows user to set :sweat_smile: For tortoisegit, the workspace follow Windows's CRLF, and inside repo index, it's LF.

sinsinpub:
Clarified the current modification of this PR: Just changed those files has different line endings inside single file, not unify the line endings of all files in the repo yet. :sweat_smile:

dragonjet:
ah i see, got it now :+1:

sinsinpub:
@dragonjet maybe no 'standard' at all. For git and Linux world users, LF is more preferred, and `autocrlf` is designed for Windows/OS X users :sweat_smile:

PR #1525, opened 2016-09-06
Webstore to Master

Crazyteitoku:
Updating kc3-translations as requested by Rephira.


dragonjet:
the X is alright in this case, proceed to re-merge

PR #1526, opened 2016-09-06
Update TL pointer

Crazyteitoku:
Updated kc3-translations as requested by Rephira.


dragonjet:
oops, we need to bump up the manifest version before merging to webstore. chrome will reject the new version if the version number is the same. just edit the `src/manifest.json` and add 1 to the last digit

sinsinpub:
I may do another PR for it if @dragonjet 's internet is still bad :sweat:

dragonjet:
my internet is fixed, its just that im in the office atm :sweat_smile:

PR #1572, opened 2016-10-30
Add sort by type in all items in SR gear page

kyoushuu:
Right now the All items in Strategy Room > Equipment List page sorts the items by Sum. However, since they come from different equipment types, the order they show up doesn't make sense, and just makes it harder to look at.

This will add a new sort type, by equipment type, in the All category. If there are equipment from the same type, they will be sorted by the default comparator for that equipment type.


dragonjet:
tested :+1: 

@kyoushuu  you may provide up to 5 links that can be added under your name in the about developers page

PR #1593, opened 2016-11-07
update expedition bonus calculation

Javran:
now with Kinu K2 and Toku Daihatsu taken into account.
there's still a slight bit of inaccuracy, but it shouldn't matter.
(see comment in code for the technical explanation)


dragonjet:
ill just put the explanation in the release notes

PR #1594, opened 2016-11-09
About page logo for contributing sites

dragonjet:


dragonjet:
@cyrismustang we're adding logos for partner sites and tools, do you have a logo (min 60x60) that we can add for your kancolle badge generator

@rephira would you like to add wgforum.kr and namu.wiki to the list? do you have a contact with them to confirm we can call them partner site? also logos for both of them if pushing though

@arialypse iirc you have a team that worked previously on the subtitles. is it the same as the [facebook page here](https://www.facebook.com/KC.TH.Translators/?fref=ts)? we can consider it partner site of "partner team" in any case.

@KC3Kai/everyone anyone else know know a contact with kancolle-db, kancolle-calc would be nice. or any other intergrations we have.

arialypse:
Yes, that page for TH team
we are agreed 👍

Javran:
just want to point out that poi-statistics have contributors from both poi (another kancolle helper) and kcwiki, translators might keep both in the title.

cyrismustang:
@dragonjet I just kinda whipped this one up real quick: http://threebards.com/kaini/badgeicon.png

Javran:
I accidentally pushed this patch to the master one while doing some minor updates, should have been reverted, sorry for causing troubles.

dragonjet:
@arialypse thanks :+1: 

@Javran ah thanks for pointing that out, is it alright if you arrange with preferred name, desc and link on the boxes for kcwiki and poi-statistics? I'm not sure on what's the preferred distinction of each.

@cyrismustang thanks, im gonna add that icon

Javran:
@dragonjet  I guess just something like "poi-statistics by kcwiki&poi" in the title will be fine.

also need to cc @Diablohu to see if he wants to put WhoCallsTheFleet into one of these sites.

dragonjet:
just an update, I am rewriting the whole about page, where we also include partners and other credits into the `developers.json` so we can easily add items later on. no more static html on file.

Issue #1345, opened 2016-05-05
Support for enemy Air Raid


LostLogia4:
Fun stuff from Tanaka.
http://puu.sh/oH0dJ/8940924783.txt

Grats to matirsadota for kindly providing us with the data.

Javran:
the data looks straightforward:

> "api_nowhps": [-1, 200, 200, -1, -1, -1, -1, 600, 500, 500, -1, -1, -1]

 then `api_air_base_attack` is the same as `api_kouku`, where damage received is in  `api_stage3.api_fdam` (damage received by ally) and `api_stage3.api_edam` (damage received by enemy)

I guess the question is how to put ally airplanes on UI. I'm not good at designing one (shrug

ReiFan49:
so to conclude Land Base related thingy,

`api_req_map/next` that contains `api_destruction_battle` performs air raid battle inside the `api_destruction_battle` data, while for any battle that have land base attack/defense related, it'll have `api_air_base_attack` ? 
also the squadron icon is `api_mst_id` right? this is differ from the [gist](https://gist.github.com/anonymous/420e9662e25ed6fe618eca1a386f9ca6) that @Javran linked at the start of even right?

LostLogia4:
Our Admiral Panel has become even more and more crowded...

mikenzb:
Hmm.. anyone noticed how battle prediction breaks when one of your squadrons get shot down fully or is removed from airbase is on cooldown?

Battle prediction works fine when your LBAS lives and doesn't get shot down or isn't removed from base and when you yasen the battle the prediction kinda works.

Attached files that might help:
[API Battle data on boss node](http://pastebin.com/raw/1wbMJ7ay)
[API Battle data in yasen on same node](http://pastebin.com/raw/17p7rii3)
![E7_Day](https://cloud.githubusercontent.com/assets/5982780/15169307/eb2bc41a-178d-11e6-8e06-d6ec6a95e1d8.PNG)
![E7_Yasen](https://cloud.githubusercontent.com/assets/5982780/15169312/f5ab7f84-178d-11e6-82c0-d83b1038849a.PNG)

Javran:
@mikenzb is #1361 related? if this is the same problem I think it should have been fixed in the latest `master`

mikenzb:
Oh Whoops, Haven't pulled on my master build in a while, yeah looks like it would be related to #1361, Let me go update my master build now.

sinsinpub:
[...]

PR #1638, opened 2016-11-28
Battle activity for Land-base Enemy Air Raid

sinsinpub:
Add processing chain as additional data attached to a node for Enemy Air Raid (in API `api_req_map/next@api_data.api_destruction_battle`) to show land-base air raid battle result in battle activity panel.
As it's in request `/next`, no persistence supported yet, will be not in sortie history.

And as it has conflicted with #1612, I've merge that first. So that will be also merged when this PR merged.

ps. panel screenshot like this:
image

### DIFF
--- a/src/library/managers/SortieManager.js
+++ b/src/library/managers/SortieManager.js
@@ -184,6 +184,7 @@ Xxxxxxx
 			// Battle Node
 			// api_event_kind = 1 (day battle)
 			// api_event_kind = 2 (start at night battle)
+			// api_event_kind = 3 (night battle first, then day battle)
 			// api_event_kind = 4 (aerial exchange)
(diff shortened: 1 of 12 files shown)


dragonjet:
this also resolves #1345?

sinsinpub:
maybe partially. just give some views also in regular map temporarily. many related features not implemented.

dragonjet:
thanks, I'm currently on E-2, I'll be able to test this out as well :+1:

dragonjet:
icon borders looks good so far :+1:

sinsinpub:
known issue about saving air raid data:
air raid data not saved when it occurs on non-battle node (eg: item get, route select)

PR #1642, opened 2016-12-02
Remote Debugging instead of DevTools

dragonjet:


dragonjet:
defer until kc3kaini. #1645 can deal with the chrome 55 problem

Issue #1643, opened 2016-12-02
[Issues] Panel don't start


nicolas22g:
I have issues with API Link.
The game start normally but the panel is stuck at "Waiting for games actions" even if I start a battle or returning to main game menu.

I already try to restart Chrome and to delete the cache. Still stuck

My Chrome version - 55.0.2883.75
My KC3 Kai version - 30.11.3

### Screenshots:
http://puu.sh/sBRPq/3b118a3d72.png

Green-Cat:
It's a chrome issue. Duplicate of #1641

dragonjet:
for this, we will be adding a "Special DMM Mode" as a new gameplay options. This customizes the DMM website directly, not enclosed in any iframe: #1645

Issue #1641, opened 2016-12-02
Chrome Version 55: Black screen


arialypse:
Chrome version - 55.0.2883.75 m (64-bit)
OS Windows 10/1511 64bit
KC3 Kai version 30.11.3

I can play game in DMM or API as usual. but  DMM Frame got Black screen

![v55](http://i.imgur.com/AAjdDgc.png)

sinsinpub:
55 totally owata/ 😭 
only dmm site works thanks to same domain. api not work for web request / network monitoring (even viewing elements inside iframe with devtools not allowed), dmm frame even black the flash player with its shit html5 stuff

dragonjet:
R.I.P. KC3 :joy_cat:

dragonjet:
the only idea I can think of is to execute the old plan where we inject styles and overlays onto the DMM website itself - so no more iframes. everyone visites DMM gameplay page directly and they get the cropped screen with custom background etc.

Javran:
sorry but I don't understand what's going on, is new version of chrome applying some new awkward mechanism of security, or dmm has figured out how to prevent user from using devtools or something? I haven't run into this problem but it seems everyone else is experiencing it...

HitomaruKonpaku:
@Javran I think its chrome problem. Try update your Chrome to ver 55

dragonjet:
on chrome 55 now on stable channel of releases, playing via DMM frame now only shows black screen. playing via API link works, but the devtools do not detect the network requests and stays in the "waiting for game actions" screen. it's just chrome problem though, no DMM intervention

HakenBrowning:
Hello, just to confirm it's not only Chrome problem. Using latest vivaldi snapshot (based on Chromium 55.0.2883.64), 2 main problems : 
- devtools not detected by game window
- when launching "Play on Frame", only white screen but sounds play too, so game only invisible.

arialypse:
I was recommended my team to use Chrome Portable for now if v54 still worked
just use separately from normal Chrome for playing KC only

http://portableapps.com/apps/internet/google_chrome_portable

DynamicSTOP:
@HakenBrowning 
> Using latest vivaldi snapshot (based on Chromium 55.0.2883.64), 2 main problems

try this:
go to chrome://flags/ 
1) disable replace flash with html
2) enable always run flash

Not sure about vivaldi but it works in Chromium 57.0.2940.0

dragonjet:
[...]

PR #1645, opened 2016-12-03
KC3改 31: Merged Gameplay Option

### LINKED ISSUE #425: Screenshot Feature Re-enabling In the Direct DMM Website Gameplay
As we all know, since the gameplay moved from DMM Frame to the Direct DMM Website, the screenshot button is rendered not functional anymore and will just disappear once you attempt to take a screenshot using it. 

This make we need to resort to our classic way of screenshot-ing the frame: Windows PrintScreen, Snipping Tool, Grab or any other similar tools. This way of taking screenshot will require a little bit more work to do, and tend to miss important scenes where timing is important, like destroying the boss bars, branching before the formation selector appears, wedding scene, etc.

Thus, this feature is somehow needed to be re-enabled in the Direct DMM Website gameplay, for instant screenshot for important scenes, as well as for easy screenshot-ing. This not need to be in next version, instead you guys can take time for more important update first before implementing this, but make sure you still implement this feature someday in future.

Cheers, guys~

### LINKED ISSUE #890: OSAPI?
I heard that except the regular API link, there are also a type of api link called as OSAPI link which have shorter lifespan and thus more secure. How about adding an option to fretch and play with osapi?

### LINKED ISSUE #1033: DMM网页模式 截图不能?
@Javran 点击development面板上的那个截图按钮, 但是却不能截图,不知道你们有没有发现

### LINKED ISSUE #1274: [Bug report] Game freezing on right click in frame
I've run across this issue a couple of times today when the game will freeze on the last screen you were on when I accidentally right-clicked on the game in the frame. This issue doesn't clear and I had to refresh the game to get to work again. This same issue also causes the Strategy Room to be inaccessible as a result.

### LINKED ISSUE #1646: Screen problem
After the new version released, my KC3 kai screen shown some problems. 

Before the new version released, it will go to dmm page first. However, it doesn't go to the page anymore now. Furthermore, the screen doesn't fit as shown in the screenshot below.

The function is still normal but not all of them. The kanmusu line didn't translate and I can't screenshot as well. 

I tried to solve these problem by remove and reinstall it, close my chrome and reopen and restart my computer but it didn't help at all.

Can someone help me?

dragonjet:
Special DMM Mode is a new gameplay option, where we direct players to the DMM KanColle play page, but we inject our on-screen customization and overlays through [content_scripts](https://developer.chrome.com/extensions/content_scripts).

This means we will not be embedding the DMM page into an iframe anymore, and the web page will be loaded naturally directly into the tab.

This move is in response to Chrome 55, which caused problems with the old gameplay options (DMM frame, and API link) where flash would not render, and/or the network requests are not detected. See more info on these two threads: #1641 #1643

----
## Related Issues
* #1641 Chrome Version 55: Black screen
* #1643 [Issues] Panel don't start
* Close #425 Screenshot Feature Re-enabling In the Direct DMM Website Gameplay
* Close #1033 DMM网页模式 截图不能?
* Close #1274 Game freezing on right click in frame
* Close #1646 Screen problem
* Close #890 OSAPI

----
## Major Changes
* Removed DMM Frame Gameplay Option
* Removed API Link Gameplay Option
  * Users also will not be able to just extract API link to use somewhere else
* Many changed name in Settings, will reset current profile's settings
  * All UI customization settings renamed
  *
  * Removed `checkLiveQuests` config
  * Removed `forceDMMLogin` config
* Removed check for DMM login. New mode goes directly to DMM so no need to check.

----
### Customization
- [x] Customizable Background Images
- [x] Cuztomizable Game Frame Margins

### Overlays
- [x] Subtitles
- [x] Quest Translation Overlay
- [x] Non-TL'ed Quest GT link
- ~~Annoying Taiha Alert~~
- [x] Map Markers Overlay

### HotKeys
- [x] <kbd>F7</kbd> Toggle keyboard docus
- [x] <kbd>F9</kbd>Trigger screenshot
- [x] <kbd>F10</kbd> Clear all current overlays

### Other
- [x] Exit confirmation
- [x] Screenshot
- [x] Fit Screen~~
- ~~Idle Timer~~
- ~~Refresher Button~~

### Integrations
- [x] New menu option
- [x] New landing page for "Play" menu option
- [x] Remove DMM Frame and API link codes
- [ ] Re-include DMM Frame and API link


sinsinpub:
rush into 'removing all old stuffs' will just cause kinds of people confusing, objecting and even panic.
a timeline should be scheduled for stable transition by marking them deprecated for now.

src/library/objects/Messengers.js
```diff
@@ -24,7 +24,7 @@ Sample Usage:
 (function(){
 	"use strict";
 	
-	console.info("KC3改 Messengers loaded");
+	console.log("KC3改 Messengers loaded");
```
sinsinpub: About this console logging... from chromium 55+, new extension box named 'collect errors' introduced, and enabled by default for unpacked extension. I guess this is the motivation u change it. But it is still collected as errors even changed to `.log` 😭  I think we may not care about this, because packed ext. and developers could both disable the 'collect errors' to avoid errors icon appear...

dragonjet: oh sorry the real motivation was that it should also be removed by the build for webstore, where only .logs are removed. I wasn't aware of the collect errors. I'm fine with .info either way. I can just configure the build to also remove .info since if that collection only works for unpacked ones, no need to keep it on webstore.

sinsinpub: ah, current grunt remove logging default config will remove levels including info, warn... that is confirmed, and why I added the mark `/*RemoveLogging:skip*/` to some critical places

PR #1668, opened 2016-12-14
Update data/edges.json

yukixz:
Using data generated from https://github.com/yukixz/kcmap.
In return for using `edges.json` as source.


sinsinpub:
kcmap SeemsGood! I'm wondering are those xml/edge IDs extracted from SWFs?
SVGs are drawn manually or not?

yukixz:
There are only two manual action:
1. Download and convert map swf to xml.
2. Mark spot name(letter).
Then kcmap/kcmap.es will extract information from xml and draw svg.

yukixz:
I wrote a post about how kcmap works.
http://blog.dazzyd.org/blog/how-to-draw-a-kancolle-map/

dragonjet:
looks good @yukixz thanks :+1:

PR #1686, opened 2016-12-23
Count number of heart-locked ships / items in devtools

Javran:
when moving over number of ships in devtools panel:
2016-12-23_133550_3840x1080_scrot

same goes with items.


sinsinpub:
`:` seems not necessary, suggest remove?

Javran:
@sinsinpub looks good, just removed.

PR #1700, opened 2016-12-30
Ship name translation method improvement

sinsinpub:
Refactor ship name meta method, allow ship names translated and combined by 'prefixes + root + suffixes' model sourced from #1696. The prefixes & suffixes (and abyssal's 'yomi') are separated into individual JSON file in TL repo named `ship_affix.json` for easier updating works.

Not only for the seasonal ship names, but also benefited for more automatic ship name (including abyssal's maybe) translation to reduce the lines of `ships.json`. For example, `軽巡ツ級` may be defined as prefix `Light Cruiser ` + root `Tsu-Class`; `駆逐イ級後期型` may be defined as prefix `Destroyer ` + root `I-Class` + suffix ` Late Model`.

For @KC3Kai/translators, the rules of affixes replacement should be known:
 * Lines of prefixes & suffixes table in `ship_affix.json` should be in correct order: 'first in first replacing', if multiply lines are intend to be applied to one ship name. Such as following:
 * For suffixes, as an example, the `甲(A)` of `瑞鶴改二甲(Zuikaku Kai Ni A)` should be replaced first, then for the left string `瑞鶴改二(Zuikaku Kai Ni)`, the `改二(Kai Ni)` becomes the real suffix. That's why in affixes table, line `甲" : " A"` should appears before `"改二" : " Kai Ni"`.
 * For prefixes, order of sub string in name to be replaced should be just reversed.
 * Don't forget the 'extending', for most languages, if some lines are not appeared in `??/ship_affix.json`, those lines in `en/ship_affix.json` will be used.
 * Also don't forget if there is a line in `ships.json` completely matched with JP name, it will be used. No replacement and combination will occur.
 * The old suffixes lines in `ships.json` will not be used any more, could be safely removed after this PR merged.

### DIFF
--- a/src/library/modules/Meta.js
+++ b/src/library/modules/Meta.js
@@ -15,6 +15,7 @@ Provides access to data on built-in JSON files
 		_exp_ship:{},
 		_gauges:{},
 		_ship:{},
+		_shipAffix:{},
 		_defeq:{},
 		_slotitem:{},
 		_equiptype:[],
@@ -89,6 +90,7 @@ Provides access to data on built-in JSON files
 			
 			// Load Translations
 			this._ship 		= KC3Translation.getJSON(repo, 'ships', true);
+			this._shipAffix	= KC3Translation.getJSON(repo, 'ship_affix', true);
 			this._slotitem	= KC3Translation.getJSON(repo, 'items', true);
 			this._equiptype	= KC3Translation.getJSON(repo, 'equiptype', true);
 			this._quests	= KC3Translation.getJSON(repo, 'quests', true);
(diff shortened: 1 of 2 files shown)


sinsinpub:
btw, the `[882] Summer 三隅` is not translated because it's official devs' typo of Mikuma 🤣  maybe add a line for it to `ships.json` 😏

Vultren:
I did notice the thing with prefixes like [Light Cruisers] and what not. Was itching to re-work the entire ships.json, but I hadn't read though the javascripts at the time and I wasn't sure how things would break.

rephira:
So to clarify the rule @sinsinpub 

The order of key-value pair present in the current ship-affix.json are in the order that they should be in order to make the system work, and hence should not be modified in any other way? (As in changing the order would influence the 2nd and 3rd point, which would cause the automation to not function as we expect?)

sinsinpub:
@rephira 2nd & 3nd points are used to describe the 1st point 'meanings of the order' additionally. So just modify them as what u do in `ships.json` as the order rule for suffixes is the same. Just notice that for prefixes, order is reversed.

rephira:
@sinsinpub Thx for clarification :)

Vultren:
@sinsinpub,  Should we have a different list for the abyssal ships or just add it to the current lists?

sinsinpub:
@Vultren just current list for now, as they are in the same list either for official master data 😃

Vultren:
Got it

Javran:
just about the `[882] Summer 三隅` typo: we can correct it directly on data received from `api_start2` instead of letting translators to make special cases in every language they maintain. (I don't even think KC devs will fix that typo soon)

ethrundr:
since it seems no more edit will be done as per now, I'll perform the merge.

Javran:
the patch looks good to me, just a slight concern: JSON does not guarantee to preserve the ordering of keys (see: http://stackoverflow.com/a/4515863, it's a java related discussion, not js, but the point still apply here), so even if translators put them in correct order, the resulting key order might still be unexpected. for now we just have few simple affixes so this problem won't get in the way.

sinsinpub:
@Javran that's true for JSON. but for `Object.keys` it still keep the order of the JSON string definition, not sure why... For current 'few suffixes', the properties order when u try to iterating in other way also changed, this is do be a tricky :joy:

c933103:
question: in scenario like "Destroyer Water Demon", is it possible to simply use "prefix: Destroyer" + "suffix: Water Demon" without a root?

PR #1706, opened 2017-01-01
Fix deckbuilder

### LINKED ISSUE #1704: [Bug] Export fleet to deckbuilder do not capture slotlevel of plane
This happened since quite a long time ago
When I export the current fleet to deckbuilder, the slotlevel of carrier plane cannot be captured and remained as 0

Besides, the content extended slot cannot be captured too

Javran:
fix #1704. refactored a bit so each object handles its own structure of deckbuilder.

### DIFF
--- a/src/library/managers/PlayerManager.js
+++ b/src/library/managers/PlayerManager.js
@@ -226,6 +226,17 @@ Does not include Ships and Gears which are managed by other Managers
 			});
 		},
 
+		// make sure to "loadFleets" before calling this function.
+		prepareDeckbuilder: function() {
+			return {
+				version: 4,
+				f1: PlayerManager.fleets[0].deckbuilder(),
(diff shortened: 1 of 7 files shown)


Javran:
update: also updated fleet manager for its deckbuilder exporting function. (history sortie won't have improvement level and aircraft proficiency because our db has no such record, we might need to deal with this problem in future, as a separated issue)

Javran:
@sinsinpub I also thought about using `.deckbuilder()`, it's possible and might even be better. but that requires more work. for the purpose of this PR, it's just catching up with latest deckbuilder format. so I think this is good enough.

I was expecting the one that pushes merge button not being the same one that starts the PR, but since I've received two approvals from you, I'll merge this one. (probably future PRs can all work in this way)

PR #1708, opened 2017-01-01
Asynchronous loading of language files

dragonjet:
As a prerequisite to loading translation files online so we do not need to do webstore releases every time there is a quest or quote update.

This might be difficult to implement now, but will try my best not to break stuff :sweat_drops: 

If this proves to take longer to figure out, we might resort to synchronous loading online translations for the mean time, although this would involve having TL versioning so we dont download it each time.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,4 +1,5 @@
 {
     "newcap": false,
-    "laxbreak": true
+    "laxbreak": true,
+    "esversion": 6
 }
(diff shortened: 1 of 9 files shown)


sinsinpub:
I'm wondering what is the file `custom` for? and is it better to continue after merging #1700 ?

dragonjet:
@sinsinpub I'm thinking custom is for data that is unique for each language, e.g. font to be used on html, file versioning for comparing webstore files vs latest files in github (while each language keeps its own version number so they can bump the update without affecting other languages).

dragonjet:
The new interface in #1722 paves a way for a new way to do translation updates without webstore. It is not yet implemented there though, I meant there will be a better approach than making Meta.js translations as async. We need to merge it first before we can start on the better approach.

Alternatively, the new "Update Page" introduced in #1748 can also be a potential place to include the translation online update.

Issue #1684, opened 2016-12-23
Suggestion: fleet AA and AACI values analizer.


Bararun:
We've recently gotten fully working AA formulas, both for our fleet and the abyssal, thanks to the info datamined from Kancolle Kai and crosschecking that info with what we knew about the browser game. You can find then in the english wikia (http://kancolle.wikia.com/wiki/Aerial_Combat#Fleet_Anti-air_Defense), japanese wikia and wikiwiki.

Since now we've got a way to acurately calculate these values, it would be a pretty great adition to KC3 if this is included in the app. Not to mention than the japanese version of Electronic Observer already has it. http://i.imgur.com/6Jr9aF4.jpg

Javran:
if this is not on anyone's todo list, I can try to implement it. but someone else still need to put it on UI as I'm generally not good at designing UI.

Javran:
self-assigned to work on the formula. 

view branch `dev-anti-air` for progress.

Javran:
all underlying works are done. see: https://github.com/KC3Kai/KC3Kai/pull/1711
any @KC3Kai/developers interested in this, please feel free to take over and I'm willing to answer related questions and adjust code accordingly.
un-assigning myself for now.

HKGPsy:
thank you for the new features
Just one question, does the fixed shotdown take into account the multiplier bonus of  AACI? What if there are multiple AACI ship?

Javran:
@HKGPsy all possible AACI setups are checked and one with the maximum benefit shows up.

HKGPsy:
Thanks for the reply , I have done some calculation with this example
[img]http://i.imgur.com/9m8jNv5.png[/img]
The extended slot contain a Bofor machine gun and its showing the info of "照月改"

The fleet AA power = [(8*0.35+3*sqrt(9))+(1*0.2)+(7*0.4)+(4*0.2)+(10*0.2) ]*2/1.3 = 26.15
The ship AA power = 114
AACI fixed shotdown bonus = 1.65 [source](http://ja.kancolle.wikia.com/wiki/%E5%AF%BE%E7%A9%BA%E3%82%AB%E3%83%83%E3%83%88%E3%82%A4%E3%83%B3)
Fixed shotdown = [(114+26.15)*1.65/10]=23

However the displayed value was 14, If I omit the AACI fixed shotdown bonus, it would be 14, so I wonder if the AACI bonus has been omitted in the current build
[...]

Issue #1683, opened 2016-12-22
In-app AS calculator


Agc0de6:
My idea is to create AS calculator as part of Strategy Room.

Two reasons behind this:
1st. English AS calculators [like this](https://a4b81641afe20619f5ed716627d72ef95dcd42d1-www.googledrive.com/host/0B37L_d6zeTfUS0puRTRFVml1czA/) are dead.
2nd. [Japanese AS calculator (original of previous one)](http://aircalc.net/) wasn't update from november. [Author](https://twitter.com/no1244/status/799786222881296384) said that he won't do this anymore.
Tanaka added too many aircraft and ships that could be used in calculator, so I think we should do this for our players' sake.

Sorry for my bad English.) As Verniy said, "Если что не так, кидайтесь тапками" ("If something is wrong, just throw slippers at me").

Agc0de6:
UI interface sample (can be changed afterwards).
Player selecting ships from lists. White fields are for slot power of every ship.
Calculator gets list of plane that player have from Aircraft tab (Aircraft Profiency and improvement bonus are excluded for now).
Then player selecting map and expecting result for this map. Orange fields are here:
1st - for max EFP of the map,
2nd - for AA buffer,
3rd - for expecting AS+/AS,
4th - for real AS+/AS.
+-D shows the difference between values in 3rd and 4th orange fields.
And then calc showing result in green fields.
*can't explain properly*
![as_calc_scheme](https://cloud.githubusercontent.com/assets/14156630/21736286/615790fe-d491-11e6-8710-6306ce6dc710.jpg)

Agc0de6:
# Requesting help for programming
*actually can't understand coding at all >.<*

ethrundr:
@Agc0de6 you can actually mention @KC3Kai/developers to ping all devs :>

Agc0de6:
@KC3Kai/developers, *requesting help for programming!*

dragonjet:
oh the existing AS calculator websites are now, for the lack of term, abandoned?

sinsinpub:
as I mentioned at [#1711](https://github.com/KC3Kai/KC3Kai/pull/1711#issuecomment-271156709), the hardest work is keeping updating the data about: every slots' capacity of all abyssal ships, encounter patterns of every nodes and every map.
I guess this is why they stopped to updating, rather than updating the algorithm of air power.

dragonjet:
[...]

PR #1711, opened 2017-01-04
fleet AA and AACI

Javran:
Implements #1684. 

- introduced new module `src/library/modules/AntiAir.js` dedicated to anti-air related calculations
- move mouse over ship icon for proportional shotdown / fixed shotdown / AACI bonus:
2017-01-04_002340_3840x1080_scrot
- fixed shotdown is related to formation (for now I'm using 1.0 as default value):
    - 1.0 for line ahead / echelon / line abreast
    - 1.2 for double line
    - 1.6 for diamond
we might consider showing fixed shotdown for all formations like:

```
Fixed shotdown:
x - y (line ahead / echelon / line abreast)
x - y (double line)
x - y (diamond)
```

which might be a bit on the verbose side.
- I leave all UI design choice to whoever willing to take over. so now i18n is delayed until we can figure out what should be put on the UI. (@Bararun would you mind giving some suggestion about what kind of info will be useful to put on UI?)
- internally `AntiAir.js` is capable of checking whether a particular type of AACI can be performed by a ship (with her equipments taken into account), I think this feature is worth pointing out and might be useful in future.

- documents that might be useful:

    - http://kancolle.wikia.com/wiki/Aerial_Combat
    - http://wikiwiki.jp/kancolle/?%B9%D2%B6%F5%C0%EF#ved2ca21


sinsinpub:
@ethrundr not ready yet, more UI changes should be applied at least.

Javran:
@sinsinpub thanks for approval and suggestions, here are my responses:

- regarding `categoryEq` thing, I think accepting more than one parameter might be a bit overkill. as most of the time we just need to test on item's category once. having multiple parameters mean we might have to use `indexOf` even if the array contains just one element. plus it's just an internal function - if we want some functional-style helpers around, we can separate that part into another module, we can then think about generalization to make them more powerful.
- for now there are no ships under type `XBB`, so I ignored it.
- combined fleet adjusted anti-air is not supported yet. I want to see whether I have implemented it correctly before dealing with combined fleets.
- I agree with updating tooltip according during battle, I guess that the best we can do; meanwhlie I think user usually keeps an eye on AA stats when they are switching equipments, something like current gun-fit hints might be better, so they can just look at the panel instead of hovering on the icon waiting for tooltips to show up. but they are too many equipments that would affect AA stats (even large caliber main guns), that conflicts with current gun-fit hints.
- extending fleet manager will be good enough I think

Javran:
@sinsinpub Yamato-class are classified as `BB`. I think `XBB` is just like `DDE`, it's been there for a while but is never used. and yes there are no carrier-specific AACIs, but general (non-sub) AACI does work on carriers.

sinsinpub:
Making larger ship library:
image

image

sinsinpub:
@Javran but there is no polymorphism for JS, u mean add more functions with different names to handle Ship  object and Ship Master?

Javran:
> @Javran but there is no polymorphism for JS, u mean add more functions with different names to handle Ship object and Ship Master?

then show me places that you need this polymorphism.

> just accessing the property with || from two classes is okay.

this *is* the check I'm talking about. why is this even necessary?

sinsinpub:
@Javran it's solved. breaking `predicate` into two parts to apparently accept master and Ship Object is the precondition of eliminating polymorphism. I wondering how to code this if it's needed.

Javran:
update: fixed a problem with fleet AA calculation.

@sinsinpub about `sortedPossibleAaciList`, I think simply comparing on `fixed` then tie break by comparing `modifier` (and finally `icon`) will do. we simply don't have enough knowledge to predict things about AACI.

also just FYI: I had some discussion on ngacn about AACI mechanism, which might have some useful info. but still I don't think that's sufficient for us to see the whole picture. so I just put down the link here:
http://bbs.ngacn.cc/read.php?tid=10692277 (Chinese)

Agc0de6:
Wait, AA calculation?
Can we use this part for #1683?
That would be a great help!
"Пожааааааалуйста!"

Javran:
@Agc0de6 not quite... AA is short for anti-air, but I think what you mean is for figuring out aircraft configurations..

Agc0de6:
@Javran ясно. Sorry for disturbing.

Javran:
looking good. @sinsinpub do you plan on adding more commits to this PR? we could list possible fixed shotdown ranges for different formations but I think for this particular PR we are pretty much done.

sinsinpub:
yes, at least I plan to add the fleet adjusted AA value to panel/Fleet Manager like air power thing. And might list up values affected by formation/AACI triggered.

After reading those wiki texts repeatedly, I notice my misunderstanding. We can only get the Formation and triggered AACI type from battle API, and there is no way for now to get these thing:
 * each individual plane slot size of abyssal ship. I max slot size database could be crafted like our replayer done, but the real slot size at stage2 (after loss of stage1 and phases before) will be hard to be guessed.
 * there is no way to know which shijpgirl is selected to do the AA fire towards enemy plane slot.
So calculation of real shot down will not be gotten real-time during battle activity, it is only able to be done as an 'AA fire simulator' at SRoom by manually inputting the enemy slot size, chosen formation, triggered AACI, so on. This should be done later in another PR.

btw, the slot size DB of abyssal ships should be the precondition for building the air power calculator related to #1683.

Javran:
are we done for this particular PR? I'd prefer to make things available in master branch asap, new features can always be added later and have another dedicated PR for it. (and it will be more convenient: for now I can't give approval to these followup commits because technically I'm the owner of this PR)

sinsinpub:
np, just let me open another PR (don't remove branch yet, still on this branch)

Javran:
as a side note, perhaps next time we could try opening PRs to `dev-xxx` branches and see how it goes.

Javran:
merging this as we've all agreed, branch `dev-anti-air` should be kept for further development.

Bararun:
This tool should at least be as good as EO's one, I'm amazed you still hadn't talked about some of the features included on it.

https://github.com/silfumus/ElectronicObserver

https://github.com/andanteyk/ElectronicObserver

Javran:
@Bararun I don't think simply giving us 2 repo links is saying anything...

Bararun:
I figured that actually seeing for yourself the set of features that include the EO's "AA Defense Detail" tool would be more useful than me simply telling you about then.

Issue #1712, opened 2017-01-04
[bug] Fleet manager fail to capture equipment slotlevel


HKGPsy:
This is particularly obvious when I export it to deckbuilder

I have a similar issue a week ago, but that one is a direct export from the game panel to deckbuilder and it only fail to capture carrier plane slotlevel. In fleet manage of strategy room, it does not capture any slotlevel of any equipment.

Javran:
duplicate of #1665 
that issue was marked as `wontfix` but I think this problem should be addressed eventually.
we can begin by updating `library/modules/Database.js` so that equipment improvement level and aircraft proficiency can be recorded in sortie history from now on... not sure whether any of @KC3Kai/developers knows how to update it?

sinsinpub:
@ethrundr no, we didn't record the api_data directly, and even not record every properties of Fleet / Ship / Gear to save space of IndexDB.

See the method in [`Fleet.js`](https://github.com/KC3Kai/KC3Kai/blob/master/src/library/objects/Fleet.js#L807), I can add `stars` and `ace` easily. And I have added new property to store LBAS info sent to sortie/defend.
But question is: is it enough for us to just add `stars` and `ace` recording?

Javran:
@sinsinpub I think LBAS confgs + item stars & aircraft proficiency should be enough.

Javran:
Fixed by #1719, #1725

Javran:
old sorties won't have improvement level & aircraft proficiency because we didn't record them. but new sorties after #1719 will.

PR #1719, opened 2017-01-07
Update recording properties of sortie IndexDB

sinsinpub:
Related to #1712. Updated `sortie` table, tested on my side.

* Add new LBAS array to record land base objects of sortied world only, and action: sortie or defend (maybe other actions also needed?)
* Add stars & ace for ship equipments including ex-slot (also fixed length array)

Next TODOs
* Fleet manager loading from sortie history could get stars & ace
* Maps/Events history could show LBAS up
* Calculate and show eLoS & air power of sortieing (the values at beginning only) for Maps/Events history

Not supports:
* eLoS and air power calculation for every Nodes at Maps History. As after each battle (even during one battle, different phase) need to be recalculated based on latest slot count and proficiency of planes, but we don't record the change of slots.

### DIFF
--- a/src/library/managers/SortieManager.js
+++ b/src/library/managers/SortieManager.js
@@ -70,6 +70,7 @@ Xxxxxxx
 				fleet4: PlayerManager.fleets[3].sortieJson(),
 				support1: this.getSupportingFleet(false),
 				support2: this.getSupportingFleet(true),
+				lbas: this.getWorldLandBases(world),
 				time: stime
 			}, function(id){
 				self.onSortie = id;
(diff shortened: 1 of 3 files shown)


Javran:
If you plan to work on fleet manager, you might try getting rid of the converted objects and just use `KC3Ship` / `KC3Fleet` instead, so `.deckbuilder()` can be re-used.

sinsinpub:
@Javran lul it's not todo list of my own, just write up the facts I already know... ofc I'm interest in it and will try to do sth. I think your works may be faster and better if you are more familiar with codes of that part.

Javran:
2 approvals should be good enough, merging this.

Javran:
@sinsinpub is `dev/sortie-db` safe for deletion? (perhaps from now on we'll delete branches after merging if not explicitly requested for keeping)

sinsinpub:
yep, np.

Issue #1690, opened 2016-12-25
Suggestion: Add in information in the expedition tabs for overdrum requirements


Tsubakura:
Heya,

The exped tab is a very handy feature, which allows players to check if they meet the requirements for an expedition or not. However, it is currently missing information regarding the requirements for overdrumming.

Overdrumming is when you equip more drums then necessary for an expedition, which grants you a chance  for Great Success. Overdrum requirement is reached when you reach 25% more than the initial required drums.
The expeditions where overdrumming is possible, are 21,24,37,38 and 40.
The following amount of drums are required for each expedition:
- Expedition 21 - 4+ drums
- Expedition 24 - 4+ drums
- Expedition 37 - 5+ drums
- Expedition 38 - 10+ drums
- Expedition 40 - 4+ drums

What I want to suggest, is to add another statement under the expeditions where overdrumming is possible. The statement is red/yellow if the requirement aren't met, just like the others.  I am suggesting yellow as an alternative color, since I also want to make it obvious that overdrumming is not required for the expedition to succeed, its only for increasing the GS chance.
Example for exp 21:
If requirements are not met for success: Overdrum: 4+(RED)
If requirements are met for success, but not for overdrum: Overdrum: 4+(YELLOW)
If requirements for overdrumming are met: 4+(GREEN)

Example for exp 24:
As Exp 24 does not require drums to succeed, the red statement doesnt exist.
If requirements are met for success, but not for overdrum: Overdrum: 4+(YELLOW)
If requirements for overdrumming are met: 4+(GREEN)

This is a way to increase awareness to newfaces that overdrumming is a thing and that exp 24 and 40 can be overdrummed as well. I would love to see this implemented. :)

Javran:
thanks for the suggestion. I'm fully aware of the fact of overdrumming, just that personally I don't like it and it might not worth the effort to implement - overdrumming can do just a tiny bit better than an all-sparkled fleet (theoretically it's just the difference between 99% and 100%), but the precondition is complicated (it's not just the number of drums, but also I believe you need some number of sparkled ships in a fleet as well) and at the time that I implemented exped tab the exact precondition was still under investigation, so I wasn't willing to put any unconfirmed info on the panel.

but anyway, I guess that's exactly what exped tab is for, maybe just a smaller indicator, or some tooltip when mouse moves over the "GS" toggle will do just fine.

Tsubakura:
That would probably be a next step actually. There are 2 things we already know:

[...]

PR #1724, opened 2017-01-08
expedition panel: great success condition checker

Javran:
Implement #1690.

Added "GS Rate" together with some hint texts:
2017-01-08_121759_3840x1080_scrot

- only visible when "great success" toggle is on
- great success rate is only available when having >=4 sparkled ships in fleet (otherwise the text is always `???` because we lack an accurate estimation for this situation)
- the text turns golden when having >=4 sparked ships
    - exceptions are expedition 21, 24, 37, 38, 40, for these expeditions we further require carrying sufficient drums before the text can turn golden.


sinsinpub:
note: remember to update TL pointer of master branch as soon as merging this.

PR #1725, opened 2017-01-08
fleet manager: reuse .deckbuilder() method

Javran:
reusing `.deckbuilder()` of KC3Fleet's


sinsinpub:
Fixed loading issue by returning empty `KC3Ship` instance from crafted `ShipManager` which is the same behavior of original one.

Loading `stars and ace` from sortie history seems really easy, just adding it, and FAQ text updated

Javran:
@sinsinpub thanks for the fix. all looks good. now together with #1719 we can close #1712 after merging both :+1:

Issue #1733, opened 2017-01-13
LBAS / Jet consumption testing request


sinsinpub:
Tests on consumption timing of fuel/ammo/steel when LBAS support/Jet assault phase triggered.
Recommended steps:

 1. Remember the resources of fuel & ammo
 2. Send LBAS sortieing to a node
 3. After entering the node battle, trigger a catbomb intendedly (by blocking network access or reload page or sth)
 4. Return to game home port, to see if fuel & ammo consumed or not, and how many it decreased (regeneration excluded would be better)
 5. Retry 1~4, change the catbomb timing of step 3. For example, after entering battle result screen, before entering home port when sortie finished, or even just after LBAS attack points selecting committed.

Similar with above, the steel consumption timing of jet bombers also needed.

btw, my assumption of that two consumption timings are both the API call of `/battle` series, not `/battleresult` or others.

LostLogia4:
Jet catbombing result (Before):
![before](https://cloud.githubusercontent.com/assets/12930396/21917447/c8253b2e-d982-11e6-989b-c3741ee4c30d.png)
Jet catbombing result (After):
![after](https://cloud.githubusercontent.com/assets/12930396/21917449/c9132cb2-d982-11e6-93c9-88dbf4ccb32d.png)

sinsinpub:
@LostLogia4 which catbomb timing? during battle, not reach battle result?

LostLogia4:
This one is catbombed after battle but before battleresult.

sinsinpub:
@LostLogia4 it seems good enough already. as consumption of steel depends on count of engaged jets, its timing must not be before battle.

Issue #1655, opened 2016-12-10
Ledger error for new records (30.12.1)


dragonjet:
still existing on 30.12.1
ledger values are 0 for new records.
as reported on facebook

LostLogia4:
Do we have any idea on the possible cause, or at least, how to investigate them?

dragonjet:
I'll ask further from the reporter

LostLogia4:
Console didn't report anything unusual. Seems like we'll need to ask people like Javran to debug this.

LostLogia4:
I've used Javran's debugging code to look up the entries and printed out a huge amount of entries:
```
KC3Database.get_lodger_data(Range(0,Infinity,0,1), function(t){t.forEach(function(b){ if(b.type.startsWith("")){console.log(b);} })});
```

And here's the list of datatype:
```
exped
sortie
quest
dsitem
critem
pvp
rmditem
crship
akashi
dsship
remodel
useitem
lbas
```
PVP appears to be an extra datatype that was unlisted in the ledger, critem and dsitem has no number suffixes

sinsinpub:
nothing useful detail offered, cannot even say there is bug. maybe just because he did not record any lbas data, maybe just no data in his selected time range 😂

LostLogia4:
Wait, you need to get new LBAS entries to fix this bug?

sinsinpub:
where is the bug? i mean there may be no any problem at all 😓 
my ledger are not 0 for new records (he mean lbas?), so what he means: 
![image](https://cloud.githubusercontent.com/assets/160240/21149394/4fcec086-c196-11e6-85f7-c3bd99e9f398.png)

sinsinpub:
write some memo on this discussion: Types of consuming not recorded by ledger for now:
 * LBAS sortie fuel and ammo consume (no api response to client found at all, amount may be calculated)
 * LBAS losses of fuel and bauxite on enemy air raid (also no api, even amount of losses not known)
 * Jet bomber steel consume (also no api, amount is being tested)

LostLogia4:
It's still all 0 on my side, and I had kept my IndexedDB since last year, so it might've been problem with my old entries. (how did you edit the Indexed DB anyways?)

sinsinpub:
😮  you mean all your data shown as 0 no matter what time range selected? hmm, it's hard to say a reason. a possible one is mass records causing array overflow. or your pc time causing wrong data recorded or  unexpected query condition...
[...]

Issue #1498, opened 2016-08-21
LBAS Support more actions


dragonjet:
- [x] Currently LBAS interface do not detect and react to resupplying bases. Upon resupply, we should reflect the new plane slot counts.
- [ ] When equipping planes to a base, show the amount of bauxite used

arialypse:
I think amount of LBAS resources used are not included in Ledger

dragonjet:
tagging @ReiFan49 just in case :sweat_smile:

PR #1734, opened 2017-01-13
Update ledger DB recording

sinsinpub:
Related to #1655, #1498, might be more.

Thanks to new game mechanism of LBAS and jet aircraft, the known fact that resources consumption for our Strategy Room ledger of these new stuffs are not correctly recorded yet. And unfortunately, these consumption are just counted at server-side, no any API data returned to client-side until player back to home port in game.

Known types of consumption not recorded for now:
 * LBAS sortie fuel and ammo, amounts are researched and verified, but may change for future events
 * LBAS losses of fuel or bauxite on enemy air raid, amount is verified but type is random
 * Jet bomber steel

Based on many verifier's hard works, and the timing confirmation of #1733, we could complete these data recording step by step.

This PR TODOs:
 * [x] Fuel and ammo consumption of sortied LBAS
 * [x] Steel consumption of Jet assault
 * ~~Possible amount of loss on enemy air raid~~ (only shown on panel activity, impossible to record). As not able to test for me, and not about DB recording, so leave it for another PR.

### DIFF
--- a/src/library/managers/SortieManager.js
+++ b/src/library/managers/SortieManager.js
@@ -375,21 +375,23 @@ Xxxxxxx
 			localStorage.setObject('sortie',this);
 		},
 		
-		endSortie :function(){
-			var
-				pvpData = JSON.parse(localStorage.statistics || "{}").pvp,
-				sentFleet = this.fleetSent,
-				self = this,
-				cons = {};
-			this.fleetSent = 1;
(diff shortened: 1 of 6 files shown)


sinsinpub:
I guess if this is not merged, no one would be able to test it luul, so mark it ready.
have tested on my side, at least no exception found for now.

Issue #1698, opened 2016-12-29
[Discussion] opinion regarding ECMAScript 6?


Javran:
in short, I'm proposing to make ES6 available into our project, but to justify the use of new features (if not obvious) developers should give detailed comments (about what the code in question is doing).

for @KC3Kai/developers, despite the fact that we already have write permissions, I recommend use PRs for non-critical bug fixes (critical fixes or patch of just few lines might still be okay), new features and massive refactorings so other developers would have a chance to take a look and demand more explanation if necessary.

I know we had some discussion on ES6 previously (#918) and the main concern was readability. 
But my point is ES6 brings some good stuff, like `x => x+1`, `let` bindings and `for (var of ary)`, if used correctly these new features can prevent namespace pollution, simplify the code (for example, always feel `$.each` is an overkill on arrays but I'm too lazy to control indices explicitly myself) and even improve readability (for example `function(x) {return x+1; }` vs. `x => x+1`; `$.each(ary, function(i,x) {...})` vs. `for (let x of ary)`)

any opinion on this?

dragonjet:
### ES6
I agree now with using ES6 since I've been exposed to a few more projects at work via nodeJS. However, I am still not familiar with its use in client-side, I mean, are we going to need preprocessors i.e. babel? NodeJS runtime already handles it but tbh I haven't made a build job before for compiling ES6.

I know chrome can support many of these stuff, but due to the way chrome updates these days, I am sure some of our players turned off chrome updates, and some of them go way back chrome 30+, or some just turned it off this chrome 54 just to avoid the black screens. and I don't know which functions will they be missing. though using a preprocessor will avoid this problem.

In general, I won't chicken away if we do agree that we need ES6, I'm fine with it and will adjust for the good of both the app and myself too. I can't stay living in the old school XD

### Pull Requests
I agree with PRs-first, unless it's a quick hotfix we need out very soon, we should probably have others take a look. I am also guilty with pushing straight to `master`, just to find out that even though it works, the goal / end result is questionable XD

In addition to this, anyone on KC3 org can review PRs, not just developers, as long as it's a peer review, having second opinion on stuff helps, not just to question the goal or the update, but also to get suggestions, quick related additions that would benefit those we didn't think of before.

Which means, you do not need to wait for my green light for a merge. As long as some are happy with it as well (would be nice if more than one person reviewed it though). We all own this now, since the time we moved the repo to the github organization XD

sinsinpub:
[...]

Issue #1730, opened 2017-01-11
[Discussion] Issue Label Reworks


ethrundr:
After looking back for the previous year regarding the issue labels, I do think we need to add/adjust few labels so it may meet its original meanings, as well as to align with our new policy regarding PR Reviews:

1. Add **type:priority** to indicate a priority issue all members should pay higher attention to and quickly solve it,
2. Rename **status:needs-correction** to **status:change-requested** to adhere with the GitHub Reviews.
3. Add **status:pending** to indicate that a current work (especially those that have been **status:WIP**-ed) for a issue/update is stalled or even dropped, so others may take over the work.
4. Rename **rework** to **status:rework**, **other** to **type:other**, **wiki** to **type:wiki** and **update_adjustment** to **type:update_adjustment**, just for standarized naming. So the only orphaned labels are **duplicate** and **wontfix**, which better to be left as is.

Also, I think it's a good practice if someone going to work on an issue reported assigns themselves to the issue. This is to make sure any other devs or reporters know who to address when testing the update or facing the same issue.

This is just my opinion tho. Any other opinion are welcomed before we actually do some changes to those labels.

dragonjet:
true, our labels are not standardized atm, gonna add my input but in list form

* `status:needs-correction` can be removed, Github reviews shows this better with "request changes"
* `status:pending`, or possibly `status:shelf`(?) but either is fine
* rework though is a type, so probably `type:rework`, others I agree

add:
*  `ready-for-merge` can also be removed, GitHub approvals does this better
* `status:testing` for me, can be named to `status:review` to let us know it's done with coding

ethrundr:
Will update the labels accordingly later at night, when my connection is better :>

PR #1742, opened 2017-01-18
Enabling ES6

Javran:
enable es6 for KC3Kai

related: #1698


Javran:
@dragonjet btw I found most of the packages are outdated:

```
 grunt                    0.4.5  →   1.0.1 
 grunt-contrib-clean     ~0.6.0  →  ~1.0.0 
 grunt-contrib-concat    ~0.5.1  →  ~1.0.1 
 grunt-contrib-copy      ~0.8.1  →  ~1.0.0 
 grunt-contrib-cssmin   ~0.13.0  →  ~1.0.2 
 grunt-contrib-htmlmin   ~0.4.0  →  ~2.1.0 
 grunt-contrib-uglify    ~0.9.2  →  ~2.0.0 
 grunt-json-minify       ~0.4.0  →  ~1.1.0 
 grunt-jsonlint          ~1.0.4  →  ~1.1.0 
 grunt-string-replace    ~1.2.0  →  ~1.3.1 
 grunt-webstore-upload  ^0.8.10  →  ^0.9.4 
 jquery                  ^2.2.1  →  ^3.1.1 
 qunitjs                ^1.22.0  →  ^2.1.1 
```

`npm-check-updates` can bump these package versions for us but I'm afraid of breaking changes and hesitated.

Javran:
could be a problem: my first attempt of fixing name shadowing problem in `src/library/modules/Translation.js` is to replace `var` by `let`, this passes jshint, but unit tests are failing:
```
>> global failure
>> Message: SyntaxError: Use of reserved word 'let' in strict mode
>> Actual: null
>> Expected: undefined
>> file:///.../KC3Kai/src/library/modules/Translation.js:138
```

I think PhantomJS (which I believe is what grunt-contrib-qunit uses for the actual js code execution) has not yet supported es6 (but is now quite close, see: https://github.com/ariya/phantomjs/issues/14506), do we need to transpile code just for the purpose of testing?

dragonjet:
we could probably instead remove `'use strict'` from the files, would that help?

Javran:
@dragonjet I think `'use strict'` keeps an eye on many other things that we could have made mistakes so maybe not..

Javran:
labelling this PR as testing, until we can find a way of testing es6 code

Javran:
all done.

@dragonjet @sinsinpub please review changes since last your approval.

changes:

- new directory `patches` for storing patch for external libs
- uglifyJS is patched to use es6-compatible version otherwise build will fail if any es6 syntax is used
- new tmp directory `build/testenv` for testing (until PhantomJS 2.5 is up I think we can do no better than this)
    - copies everything necessary from codebase
    - transpiles non-external js code so that PhantomJS can run
    - transpiling happens only for tests, a direct build (through `grunt build`) still preserves es6 code
- some extra dependencies for all mentioned above to happen

if someone knows how to tell qunit to use `build/testenv` as root directory for testing, please let me know or commit directly on this branch. (but for now this is just a minor issue, and I think current PR is still good enough for merging.

say welcome to es6 (or, more formally, ECMAScript® 2015)! xD

dragonjet:
ah this is the first time I've seen git patches in action.

dont we manually specify in test html file which scripts to import, so we could probably import those from `build/testenv` instead of `../../../src/`

welcome es2015!

Javran:
@dragonjet what about using something like `${testenv}` for specifying the root and in grunt file we could probably replace it with real root path?

Javran:
@dragonjet update: I just tried using absolute path and it doesn't work. not sure what is supposed to be the root directory for PhantomJS, but it doesn't seem to be project root. anyway as long as current relative path works, I'm not worried too much.

Javran:
there are changes since last approvals, but I'm not receiving many comments or approvals on them.
this change could potentially be big so I need more @KC3Kai/developers 's opinion before I feel it's ok to push the merge button.

dragonjet:
I will switch and take a look at this after I finish working on the other branch

dragonjet:
I'm getting this error though
```
>> Uglifying source build/tmp/library/modules/Translation.js failed.
Warning: Uglification failed.
SyntaxError: Unexpected token: name (errMsg).
Line 133 in build/tmp/library/modules/Translation.js
 Use --force to continue.

Aborted due to warnings.
```

runtime:
* node 4.2.1
* npm 2.14.7

Javran:
@dragonjet have you applied the patch? see before_install part of .travis.yml

also note that you need to "npm install" twice, the second time is after applying the patch, for new uglifyJS to be pulled in.

dragonjet:
ah I forgot about the patch, I thought it was just for travis XD
I will rebuild tonight or tomorrow morning, I'm away from my main computer now

dragonjet:
although I had problems applying the patch, building looks good

approved again :+1:

dragonjet:
note: I did not untag the `status:merge-ready` label. I removed the label entirely from repo. Github approvals should work better #1730

Javran:
@sinsinpub any comment on this?

for all @KC3Kai/developers: 
let's allow a ~12 hr window for comments because I really want to hear what other might think about this.
after merging this PR, we can think about ES6 syntax when working with KC3Kai, but readability is always a priority despite we now have access to more fancy stuff.

sinsinpub:
it's okay for me as long as npm & grunt building can be passed perfectly.
not able to test for me as even not have npm installed on my mobile devices...
will more concern at runtime

Javran:
I guess I need to write this down somewhere in case others are experiencing the same problem as @dragonjet did.

in order to build this branch, you need to execute following lines:

```
# npm install -g grunt-cli
# npm install
# patch -d node_modules/grunt-contrib-uglify -p2 < patches/uglify-harmony.patch
# npm install
```

the `patch` command patches `grunt-contrib-uglify` to make it depend on an ES6 compatible version of uglifyJS. and because of this the second `npm install` is needed.
you can also read [`before_install` part of .travis.yml](https://github.com/KC3Kai/KC3Kai/blob/3ae7e489e99b1ed0a86f816ee5b287163a496261/.travis.yml#L4-L7), it lacks the last `npm install` command, which is the default action for travis's install step.

Javran:
@sinsinpub what do you mean by runtime, es6 performance or chrome compatibility?

PR #1744, opened 2017-01-18
Release/30.15.2

dragonjet:


dragonjet:
this proves the build problem is not on our side but on travis. will do a manual webstore upload for now

PR #1756, opened 2017-01-24
List up recent news feed

sinsinpub:
This [old designed feature](https://github.com/KC3Kai/KC3Kai/blob/857f35e984bd50e136e70f3c55a033d15b68b879/src/pages/strategy/tabs/TabHome.js) is requested again... XD
Limited by API data, only the recent logs of news feed will be saved into local storage since last time back to home port in game.
Then list them up at Profile page of Strategy Room. Support translating messages into English based on type ID.

Profile SS:
image
Homeport SS:
image
(so why are there 6 lines in API? dunno... 😂 )


mikenzb:
<3 Thanks Sin

PR #1760, opened 2017-01-25
Hotfix 30.16.1

sinsinpub:
⚓️  Urgent fix missing file in webstore build which mess up DMM frame / API mode

⚓️  Other minor fixes

 * Fix checking of LBAS holding item
 * Fix dark theme color of select/input control

⚓️  Update translations

btw: it's time to check if your extension is forced to auto-restart or not 😃

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -191,7 +191,7 @@ module.exports = function(grunt) {
 							replacement: ''
 						},
 						{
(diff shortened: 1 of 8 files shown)


sinsinpub:
dj will not be free until tomorrow... i have no npm/grunt environment to test...
so pray this patch will work  :joy:

PR #1767, opened 2017-01-29
Fix battle rank

### LINKED ISSUE #1766: wrong rank prediction result
replay http://i.imgur.com/Db3230b.png

- predicted result: B rank
- actual result: C rank

Javran:
fix #1766 

most of the discussions are in the issue above too.


Javran:
@sinsinpub did u say somewhere that dj is not around? if so I guess there are not many active devs so we can go ahead and merge.

sinsinpub:
hah, not sure, his recent works usually on weekend or midnight. just go ahead for the fix

PR #1768, opened 2017-01-29
icon for Arashio Kai Ni

Javran:
icon for Arashio Kai Ni


sinsinpub:
what should I review on this ... nice adult woman face? lol

Javran:
@sinsinpub now I don't really know why should I do this as a PR anymore.  for me `review` tag is like to say that I'm done and the PR in question is ready to land. `merge-ready` is gone so this is the closest one I can think of.

Javran:
come to think of it, why don't we automate `src/data/icons.json`?

sinsinpub:
how to...? this id list seems just standing for the existence of icon files, so autogen it means scanning icon files?

Javran:
@sinsinpub yes, that's exactly what I mean. I believe grunt should have some plugin that does the job, haven't investigated though.

sinsinpub:
seems not acceptable solution for me... a generated data file depending on grunt task will make dev version not directly available without running grunt cli once for each time icon files added.

( :trollface: : npm install eat 1GB space, it's not lite for only 1 project only take less than 20MB)

Javran:
@sinsinpub alternatively we can just put a script file somewhere and only run it when new icon gets added. we don't need to do this every time.

PR #1769, opened 2017-01-30
tool for generating icons.json

Javran:
well I'm not sure if you guys like this idea

- `update_icons.pl` for manually updating `src/data/icons.json` after adding new icons
- `patches/` -> `misc/` for storing things not supposed to be used by user but might be helpful for development


sinsinpub:
looks fine. how about merge the folder with `files` which also dev only files but not chrome extension sources?

Javran:
I have no idea what `files` is for though, if they serve the same purpose, I'll definitely consider merging into that directory instead.

sinsinpub:
they are all `.fla` in `files`, the sources of flash ship card viewer, icons (vector image) designed by dj.

Javran:
@sinsinpub let's just stay with `misc` then ... since there are all `.fla`s in `files`, having files with other extensions in it would feel like out of place.

PR #1772, opened 2017-02-01
Expedition Table and Configuration

Javran:
basically an expedition table to see resource incomes. but it's a bit hard to explain what exactly this is, I recommend you guys have a try on this branch.

it's currently located in `Dev-Only > Table`, as for now the corresponding feature in Expedition Scorer is not yet implemented. but this will be moved to `Expeditions > Table` in future.

see https://github.com/KC3Kai/KC3Kai/blob/7a8a1ccef9b0ee6665522b512e93eda02da20bb8/src/pages/strategy/tabs/expedtable/qa.md for detailed explanation (also included in Q & A section of this tab).

again I can't give a timeline about when I'll update expedition scorer accordingly. This PR is just for landing this SR tab early, making it available (currently only for developers) and looking for feedback. (so don't remove the branch after merging)


sinsinpub:
😮 i'm sure should do a git squash (`git rebase -i` or `--autosquash`) to merge such a large commit list...

Javran:
@sinsinpub I certainly know that, but I prefer if every commit is a sensible patch by itself..

dragonjet:
agreed with @Javran iirc I even removed squashing on repo settings

sinsinpub:
cannot agree. readability also applied to commits history. the intermediate process may make sense for yourself, but not for others/reviewers. others will care about a whole, final patch itself, not every modifications in history.
many commits in one purpose, temporary commits, reverted changes, cleanup, these kind of commits definitely should be squash into one. so my point is not squash all commits into one, i mean group nearby commits to many groups, see if they should be one, just squash them.

dragonjet:
hmm I see it now from both perspectives. So our possible compromise here would be,  I will enable **squash** and **rebase** options in the repo settings. However, we cannot force everyone to do one way or the other. imho as long as we make pull request descriptions understandable, I'm fine with whatever commit count there is. For me though, merge-rebase is more acceptable in concept but I haven't tried it myself

sinsinpub:
haha, i don't mean to use the squash/rebase of github. i'm just suggesting do some 'clean up' of git commits for better history readability if committer feels necessary. no one should be forced.

scenes for git squash/amend:
for example, if a commit for adding missing semicolon `;` or a modification to satisfy jshint, could be squashed or amended to previous commit, if u r not intend to left it to educate others.

another example, there are many commits with simple comment across several days, after they are merged into master, many other commits during these days will be among them. Long time later, one of minor change of those intermediate commits is hard to understand its purpose, until finally find out which PR it belongs to.

Javran:
@sinsinpub you do have a point about grouping neighboring commits if closely related. so yes I'll do that for not yet pushed commits.

For all PR reviewers: I want to restate that this PR is just for adding this page into Dev-Only pages. For now all work on this page is done and ready for reviewing. Further work might be continued but I don't have a timeline for it. as it sits in Dev-Only part of SR (for now), I don't think it would affect normal users.

sinsinpub:
have a try, looks good.
two suggestions about UI:
 * add `line-height: 20px;` to `.tab_expedtable .exped_row .exped_view .info_col.edit_btn`, triangle button will look better
 * any change on `View` bar will make selected `Sort` class disappear. beside, if there is `Reverse order`, add an indicator like ⬆️ or ⬇️  to selected order button will be good

Javran:
@sinsinpub thanks for the feedback, will adjust accordingly. and about the sort button being de-selected, it's actually intended, as switching views, especially between total / hourly views is likely to break the sorting order, having them still selected might confuse users.

alternatively we can keep sort buttons active, and redo sorting everytime a view is switched, but since every expedition row is a config by itself, I want to keep every row in their places as much as possible, so they don't get shuffled by simply changing views.

Javran:
haven't got time continue working on this yet, I'll make few small improvement & fixes and them merge this (perhaps still need at least one approval on incoming changes, just to make sure someone other than me has took a look at this branch):

- [ ] `line-height` styling
- [ ] order indicator
- [ ] about sorter being de-selected, probably will explain that in Q & A part
- [ ] current "hourly" view actually shows resource gain per minute, need to multiply it by 60 and see if we need some adjustment to UI..

Javran:
now this should be everything I've planned for this PR, for people that have approved this PR please review new changes.

Javran:
sorry for the messy commit log, I'm also annoyed by this but for now I don't have a better solution but to check https://github.com/KC3Kai/KC3Kai/compare/master...dev-sr-exped

dragonjet:
Did some interface updates at `dev-sr-exped-ui` for dark theme. You can merge that PR to this branch if it's alright.

Javran:
switching to labal WIP for now, I notice the layout is broken for en, changing to a smaller font size seems to fix the problem. I need to double check on things before I feel confident to merge.

Javran:
merging this just as I said earlier

PR #1778, opened 2017-02-01
fix "off by one" error for exp calculation

### LINKED ISSUE #1661: Off by one error in leveling goals in panel
Since the experience bar updates after the battle has ended but the "battles left until goal" number shows up before this, the "battles left until goal" is off-by-one.

image

1-5 FS+MVP gives 540 exp, ceil(4752/540) is indeed 9 (goal is level 60). However, the experience bar has not yet updated to reflect the 540 experience that was gained during this battle.

image

So the number that should be displayed should always be one less than is displayed.

Not sure if this was intentional or not. It's not really a big problem, it just sort of bugged me.

Javran:
Fix #1661

- also with some proper variable scoping.
- plain theme doesn't seem to actually using this feature, but the code
  is exactly the same, so I have the same fix applied anyway
- at some point in future we should consider merging duplicated codes

### DIFF
--- a/src/pages/devtools/themes/natsuiro/natsuiro.js
+++ b/src/pages/devtools/themes/natsuiro/natsuiro.js
@@ -40,8 +40,7 @@
 	var uiTimerLastUpdated = 0;
 	
 	// Experience Calculation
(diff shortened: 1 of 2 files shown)


sinsinpub:
btw, just see codes work properly, not confirm it fix issue or not 😃

PR #1780, opened 2017-02-02
Settings to disable Taiha alert

### LINKED ISSUE #1480: Disable Taiha Alert on Home Port
Settings so that we can enable/disable Taiha alert when not on sortie, because it doesnt matter when in Home Port

sinsinpub:
Close #1480 this time.

Two new settings added:

 * Disabled at Home Port, default: false. Enable it at player's own risk. Because it's one of two situations which should alert player NOT to go ahead with Taiha ship.
 * Ignore Damecon Ship (in other world: Disabled if Damecon equipped), default: false

can not find any reason to show Taiha alert on PvP battle... so removed the unused term of TL.
I guess it was supposed to alert at Home Port to stop player sortieing for PvP if there is Taiha non-flagship.
(i'm wondering what is the English term to describe 'non-flagship ship' which referred to `僚艦` or `随伴艦` 🤔 )

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -480,7 +480,7 @@
 			},
 			{
 				"id" : "alert_taiha_damecon",
-				"name" : "SettingsDisableHasDameon",
+				"name" : "SettingsDisableHasDamecon",
 				"type" : "check",
 				"options" : {}
 			},
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -473,6 +473,18 @@
 				"options" : {}
 			},
 			{
+				"id" : "alert_taiha_homeport",
+				"name" : "SettingsDisableHomePort",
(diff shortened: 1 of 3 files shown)


nonsensed:
I tried to find a naval term commonly used for non-flagships. There isn't one that I can find.
僚艦 = Consort Ship
随伴艦 = Accompanying Ship

As far as I know, neither works in every situation for a non-flagship. Consort and accompanying are just general English words to describe this, I believe. "Accompanying" can also be used to describe a specific type of escorting (which, itself, can mean other naval terms), so its meaning can be different from "non-flagship".

To keep it simple for players, I would just use "Non-Flagship", personally, because this is very specific for what you're using. Otherwise, it's time to find someone specialized in English ww2 naval history.

PR #1784, opened 2017-02-04
Expedition Scorer Improvement

### LINKED ISSUE #792: Suggestions for expedition scorer
Related #722, few suggestions about the scorer.
## Select by expedition time

Like what we have in the ship list:

2015-09-08_205616_3840x1080_scrot

We can make a similar selector for selecting expeditions that takes less than `x` hours (`x` could be 7 / 12 / 24)

A more flexible way would be to have a text box for user to input x.
## Great success

Run expeditions with all-sparkling fleet will have an almost guaranteed great success,
which could have an impact on the result table. I'm thinking about having another checkbox for each expedition, if user can form a sparkling fleet for that expedition, he can check it and the gross income will be multiplied by `1.5` when doing calculation.

For now I haven't implemented this, but this will be the very next feature in my lib.
## User experience
- Remember expedition selections, AFK time, etc.
- Like ship list, highlight current item as mouse moves over
- For expedition list, it is possible to click on the text to toggle checkbox?

### LINKED ISSUE #1530: Improvement for Expedition Scorer
Hello

I have a little improvement to propose (I use this to calculate the best expeditions to do):

In the Expedition Scorer, there is already a nice function that allows to get the best expeditions by giving values to the different ressources.
It would be a good thing to add the same kind of function that would use the current ressource ammounts to get the best expeditions to go back to an even level of ressources (ie. 1/1/1/1).
This can be done by calculating the reciprocal of the amount of each ressource (or 100000 times this value for practical reasons).
For exemple, if I have 15000 fuel, 10000 ammo, 8000 steel and 5000 bauxite, the values would be 
- fuel = 100,000/15,000 = 6.66
- ammo = 100,000/10,000 = 10
- steel = 100,000/8,000 = 12.5
- bauxite = 100,000/5000 = 20

This could be calculated automatically and this function could be enabled with a check box near the Resource Priority input boxes.

Javran:
while #1772 is up for reviewing, I'll start working on expedition scorer on top of it. this PR is for keeping track of update plans.

- closes  #792
- closes #1530

# planned

- [x] a separated module for functions related to expedConfig
- [x] rework on actual sorting algorithm
- [x] too many changes, reworking on UI
- [x] save user selection
- [x] change resource priority to use sliders
- [x] add button "balanced" and "reset" for resource sliders for easier reset / auto priority. however "balanced" will be hidden for now. as solving #1530 might be delayed
- [x] both themes
    - [x] legacy
    - [x] dark
- [x] update Q & A
    - [x] "Balanced" button
    - [x] remind user that the top one might not always be the best
- [x] hide result part if we have not yet calculated anything
- [x] remind user if the config and result is out-of-sync
- [x] try to address #1530 
    - "auto" button will do that
    - let's just keep things simple: take the reciprocal and scale. that'll be it. but instead of naming it "auto", let's just say it's "balanced".

# delayed

content moved to #1798


Javran:
update: most of the planned features have been completed, everyone is welcomed to have a try. 

- basically this is still the old Expedition Scorer with some slight changes and improvements.
- you don't have to configure Expedition Table for this to work, but having Expedition Table configured will take more of your expedition preference into account.
- colors are tuned just for dark theme, but this is still workable in legacy theme. (but legacy theme is in my TODO list for this PR)
- I've splitted my TODO into 2 parts: "planned" for things I'm going to complete for this PR, "delayed" for non-urgent features and things that I need more time to think about. maybe I'll open an issue and move the delayed list there.

Javran:
update: I've done every feature planed for this PR. there will be some more actions like picking commits from other branches and formally moving "Expedition Table" to "Expeditions" category, but there will be nothing important in terms of feature. so this PR is ready for review.

Javran:
sorry about this messy commit log, I thought github would arrange merged commits in a more reasonable way.

see difference between `dev-sr-exped` and `dev-sr-expedscorer` for a more clear view:
https://github.com/KC3Kai/KC3Kai/compare/dev-sr-exped...dev-sr-expedscorer

dragonjet:
hmm might wanna update target branch to master?

Javran:
okay, for now the plan is: let's worry about #1772 first. it only needs review on newly added few commits. after that I'll try to clean up the mess in this branch

Issue #1717, opened 2017-01-06
[Suggestion] dynamic min-max value for resource graphs


Javran:
Resource graphs are not informative when resources are stockpiled exceeding certain amount.
This is my per day graph, these are almost straight lines and I can barely see the difference:

![2017-01-06_183130_3840x1080_scrot](https://cloud.githubusercontent.com/assets/1096354/21736398/6ae35d18-d43e-11e6-8290-f2ade7e0cac0.jpg)

my suggestion is to scale y-axis based on min and max value of the data we are rendering.

as a side question, what is the version of `Chart.js` we are using? I think it's outdated and perhaps we should bump its version and adjust the code accordingly. I want to make improvement related to resource graphs but sadly can't find the corresponding doc to refer to.

Amelek:
Maybe it would be possible to create a relative graph? Even with dynamic scale, it won't help for items - for example, 200 screws vs 3000 instant construction relative graph will inevitably decay to current unreadable state

Also, it is a technical limitation that only single month is displayed? Given that kancolle works in monthly cycle due to extra operation reset, it would be great if we could see few more iterations (like 3 months) to compare the clear performance.

sinsinpub:
seems it's Chart.js-1.0.x... no docs to be found online 😭

Javran:
@sinsinpub yep, I used to think that a popular lib without a straightforward way of viewing past-version document was unbelievable...

Javran:
not sure how the actual range is computed, but the graph becomes funny when you have stockpiled enough resources:

![2017-01-30_165102_3840x1080_scrot](https://cloud.githubusercontent.com/assets/1096354/22443278/5ac93124-e70c-11e6-99ed-9c9ddda2ffc1.jpg)

dragonjet:
`src\pages\strategy\graphable.js`
we just need to make this false
`Chart.defaults.global.scaleBeginAtZero = true;`

Javran:
@dragonjet just tried and that does look better, any concerns about setting it to `false` previously ?

dragonjet:
there's no problem with setting it to false, but we might want to make it configurable, though I wouldn't recommend adding to global settings page since this is a view-only configuration that affects 2 pages

Javran:
@dragonjet I think putting a toggle on UI will do. also we might need to bump chart.js first because the current one we are using is outdated and kind of unmaintainable as there's no way to find the document for this version..

dragonjet:
[...]

Issue #420, opened 2015-07-27
Resource Graphs: Filter date range


ethrundr:
If possible, I'd like to suggest a feature where the user is able to select custom duration to see their resource and consumables growth during that period. This will be helpful to plan the expeditions and preparing especially for events.

This idea comes from the Facebook Page's Insight, which allow the page admins to change the timeframe included in the Insight, so they can tell the page growth for specific timeframe.

Issue #75, opened 2015-05-13
Resource Graph: Count difference


dragonjet:
_Instead of displaying each resource count at certain times, could there also be a option to see the resources lost or gained within a given time period without me manually calculating?_ - Epicureanpancake

_Hi Dragonjet, i'd like to make a suggestion about the resource graphic in KC3 Kai,i think it would be pretty userful if you made a graphic that shows how much you got or lose after one day, different from the others,this one start with 0 and show how much resources you got or lose in one day.
Well i'm asking this because if you see this graphic of mine you will see something like a straight line and this does not help at all,so it would be great if you make a "zoom out" version of this._ - [Failnaught ](http://kancolle.wikia.com/wiki/Thread:289617)

Issue #73, opened 2015-05-13
Resource Graphs: Toggle resource types


dragonjet:
requested by Epicureanpancake

as further explained by Flonnaru, checkboxes add or remove resource lines from the graph to change axis-scaling.

additionally, a checkbox "Start with zero", which can be disable to start the y-axis near the minimum.

PR #1785, opened 2017-02-04
Resource and Consumable Graphs Enhancement

dragonjet:
An enhancement to Resource and Consumable tabs in Strategy Room
* [x] #1717 Ability to make the graph not start at zero value (useful for high quantities)
* [x] #420 Graph filter date range
* [x] #75 Ability to  show difference(deltas) instead of absolute values in graph
* [x] #73 Toggle resource types in graph
* [x] Update `chart.js`
* [x] Supply sample presets

### Screenshots
chrome_2017-02-11_07-50-00

### Notes
* This also introduces the use of web workers. These are used best when there are synchronous scripts you need to execute without making the frontend unresponsive. Web workers execute in a separate thread.

### Bugs
* [x] Shows two data sets in tooltip when points are too close
* [x] Support typing on the calendar input
* [x] As of @Javran's image, the last item in graph on delta is 300k / value of absolute
* [x] Clicking X on a date input which clears the value stucks at "Loading..."


src/pages/strategy/graphable.js
```diff
@@ -1,165 +1,149 @@
...
-			var self = this;
-			return {
-				labels: data.hours,
-				datasets: Array.apply(null,{length:Object.keys(itemData.name).length}).map(function(c,i){
-					return self.dataset(itemData.full[i], itemData.colorbyte[i], itemData.colorhex[i], data[itemData.name[i]]);
-				}).reverse(),
-			};
-		};
+		// New chart JS only accepts an array, not object
+		data.datasets = Object.values(data.datasets);
```
sinsinpub: NOTE: `Object.values` is ES7 feature, which supported from chrome 54.
see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values

PR #1787, opened 2017-02-05
Configurable panel box background

### LINKED ISSUE #1786: Suggestion: Option to modify dashboard's left box's opacity
It would be a nice QOL change to allow it's opacity to be modified to block less of the background.
2017-02-04 23_20_17-kc3

dragonjet:
Close #1786

to be included in 30.17.1 - #1788


sinsinpub:
default value not look the same with before... tab text hard to see

dragonjet:
actually this went through alot XD
first attempted making 3 settings, background, fore color, opacity. removed opacity since it can be configured in back color setting by using `rgba`

attempted making text color automatic invert of background. efforts didn't work. tried `mix-blend-mode` and some other weird CSS I saw,

attempted just making text color also configurable. but the problem with custom font color is that we have multiple color highlights like red and green ones. they would have to configure it as well.

eventually removed everything else except background configuration.

ill try to make it tab text colors static, and the general recommendation is that don't stray too far from the original dark color for best results. just alter the shade of black XD

Issue #1693, opened 2016-12-26
Suggestion: Strategy Room Screenshot


HitomaruKonpaku:
Hi

Can you guys add the screenshot button to the Strategy Room to capture full page? Maybe under Back to Top button.

Sometime I need to share my ship list with my friend but it take to long to take a pic and merge them into 1 pic

sinsinpub:
recommend a screenshot tool for Windows: PicPick
a ship list exporting feature is under construction by @DynamicSTOP

HitomaruKonpaku:
@sinsinpub since my screen resolution is 1366 x 768, the page is very long, i wonder that tool can screenshot full page?

Also I think about we can screenshot any page in Strategy Room, not only ship list

DynamicSTOP:
@HitomaruKonpaku i have tried couple screenshot plugins but it looks like they work only with actual web pages, not plugin pages.
Something like [this](https://gist.github.com/DynamicSTOP/ae72262d1a20d1951df4269adffaf99c) might work. For the current downsides:
- it does not take into account width
- it does not take into account zoom
- it does not tested on hdpi

As for my plans i was planning to do something much more simpler then we have on strategy room page.

HitomaruKonpaku:
@DynamicSTOP ya the plugins not work with chrome page, that is my problem.

And thanks for your js file, it works pretty good. Also I think [this](http://html2canvas.hertzen.com/examples.html) could help too.

Btw: Hope to see your thing soon xD Thanks again.

dragonjet:
#1792 has option to export shiplist to an image in one click

Issue #1566, opened 2016-10-28
Showcase improvement


Javran:
First I have no plan working on this, just an idea, if any @KC3Kai/developers are interested, feel free to take it over:

Inspired by [this reddit thread](https://www.reddit.com/r/kancolle/comments/58c8im/discussion_fall_2016_event_preparations_thread/). I think it might come in handy if we can allow user to export / import a list of:
- all heart-locked ships (including level, stats)
- all locked equipments (including stars + air proficiency)
- all resources

the save file might be just in JSON format or encoded in some image file.

the purpose of this is to make it more convenient for people to show off or look for advice on event preparation.

this functionality overlaps largely with "Showcase" tab, I think it's a good chance to overhaul it and make it more useful, therefore the title.

PR #1792, opened 2017-02-07
new ship\equip export tool on showcase page

DynamicSTOP:
Two new buttons inside showcase tab. http://imgur.com/a/67gxg

Export are made in "download only" mode, because i am not sure that chrome will be ok about long data url. Though forcing it to use screenshot setting (imgur\download\new tab) or adding new one into options shouldn't be a big problem.

Would be really glad if someone could:
1. review code and point places that i should change (i am more like php developer and scopes in js makes me crasy)
2. test equip export. i am not even 1yo teitoku so i don't have a lot of rare stuff.

### DIFF
--- a/src/library/objects/ShowcaseExporter.js
+++ b/src/library/objects/ShowcaseExporter.js
@@ -1,0 +1,875 @@
+"use strict";
+
+var enableShelfTimer = false;
+
+function ShowcaseExporter() {
+    this.canvas = {};
+    this.ctx = {};
+    this.allShipGroups = {};
+    this.loading = 0;
+    this.shipCount = 0;
+    this.rowParams = {
+        width: 330,
+        height: 35
+    };
+    this.colors = {
+        odd: "#e0e0e0",
+        even: "#f2f2f2"
+    };
+    this.columnCount = 5;
+    this.addName = false;
+    this.loadingCount = 0;
+    this.callback = function () {
+    };
+    this.aircraftTypes = [];
+    this._statsImages = {
+        fp: null,
+        tp: null,
+        aa: null,
+        ar: null,
+        as: null,
+        ev: null,
+        ls: null,
+        dv: null,
+        ht: null,
+        rn: null,
+        or: null
+    };
+    this._equipTypeImages = {};
+    this._shipImages = {};
+    this._otherImages = {};
+    this._equipCanvases = {};
+    this._equipGroupCanvases = {}
+    this._paramToApi = {
+        fp: "api_houg",
+        tp: "api_raig",
+        aa: "api_tyku",
+        ar: "api_souk",
+        as: "api_tais",
+        ev: "api_houk",
+        ls: "api_saku",
+        dv: "api_baku",
+        ht: "api_houm",
+        rn: "api_leng",
+        or: "api_distance"
+    };
+}
+
+ShowcaseExporter.prototype._init = function () {
+    this.canvas = document.createElement("CANVAS");
+    this.ctx = this.canvas.getContext("2d");
+    this.allShipGroups = {};
+    this.aircraftTypes = $.unique($.merge(KC3GearManager.carrierBasedAircraftType3Ids, KC3GearManager.landBasedAircraftType3Ids));
+    for (var i in KC3Meta._stype) {
+        if (KC3Meta._stype != "")
+            this.allShipGroups[i] = [];
+    }
+};
+
+ShowcaseExporter.prototype.complete = function () {
+    (this.callback || function () {
+    })();
+};
+
+ShowcaseExporter.prototype._drawBorders = function (canvas, ctx, rowWidth) {
+    if (!canvas)
+        canvas = this.canvas;
+    if (!ctx)
+        ctx = this.ctx;
+    if (!rowWidth)
+        rowWidth = this.rowParams.width;
+    for (var x = rowWidth; x < canvas.width; x += rowWidth) {
+        ctx.fillStyle = "#FFF";
+        for (var y = 0; y < canvas.height; y += 20) {
+            ctx.fillRect(x - 1, y, 2, 10);
+        }
+    }
+};
+
+ShowcaseExporter.prototype._addCredits = function (canvasData, topLine) {
+    if (!canvasData)
+        canvasData = this.canvas;
+    if (!topLine)
+        topLine = "Ship List";
+
+
+    var canvas = document.createElement("CANVAS");
+    var ctx = canvas.getContext("2d");
+
+    canvas.height = canvasData.height + this.rowParams.height * 2 + 18;
+    canvas.width = canvasData.width;
+
+    ctx.fillStyle = "#FFF";
+    ctx.fillRect(0, 0, canvas.width, canvas.height);
+    ctx.drawImage(canvasData, 0, this.rowParams.height * 2, canvasData.width, canvasData.height);
+
+    var d = new Date();
+
+    var monthNames = ["Jan", "Feb", "Mar", "Apr", "May", "June",
+        "July", "Aug", "Sep", "Oct", "Nov", "Dec"
+    ];
+
+    var created = "Made in KC3 on " + d.getDate() + " " + monthNames[d.getMonth()] + " " + d.getFullYear();
+    ctx.font = "400 14pt \"Helvetica Neue\", Helvetica, Arial, sans-serif";
+    ctx.fillStyle = "#000";
+
+    ctx.fillText(
+        created,
+        canvas.width - this.rowParams.height * 0.5 - ctx.measureText(created).width,
+        canvas.height - 1
+    );
+    var x = 20;
+    if (topLine.indexOf("Ship") != -1) {
+        x = this._addConsumableImage(ctx, canvas, x, "medals") + 50;
+        this._addConsumableImage(ctx, canvas, x, "blueprints");
+    } else {
+        x = this._addConsumableImage(ctx, canvas, x, "devmats") + 50;
+        this._addConsumableImage(ctx, canvas, x, "screws");
+    }
+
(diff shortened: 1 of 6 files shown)


DynamicSTOP:
my biggest concern would be [this](https://github.com/DynamicSTOP/KC3Kai/blob/showcaseExporter/src/library/objects/ShowcaseExporter.js#L824) function. It was extracted by IDE, so i am not sure is it ok or not.
It used to draw most valuable stat first e.g. fire power for guns.

dragonjet:
I don't see anything wrong with it. Looks like its properly packaged into a single module so I don't think there will be problems with integration with the rest of the code.

src/library/objects/ShowcaseExporter.js
```diff
@@ -0,0 +1,881 @@
...
+        ctx.fillRect(0, 0, canvas.width, canvas.height);
+        ctx.drawImage(canvasData, 0, this.rowParams.height * 2, canvasData.width, canvasData.height);
+
+        var d = new Date();
+
+        var monthNames = ["Jan", "Feb", "Mar", "Apr", "May", "June",
+            "July", "Aug", "Sep", "Oct", "Nov", "Dec"
+        ];
+
+        var created = "Made in KC3 on " + d.getDate() + " " + monthNames[d.getMonth()] + " " + d.getFullYear();
```
sinsinpub: hint: date can be formatted with `format` method, see `global.js`

src/library/objects/ShowcaseExporter.js
```diff
@@ -0,0 +1,881 @@
...
+            else
+                fontSize = 10;
+            if (equip["s" + i] === 0)
+                continue;
+
+            var text = "";
+            if (i === 0) {
+                text = " x";
+            } else if (i === 10) {
+                text = "+10★ x";
```
sinsinpub: ingame it shown as ★+10, maybe. btw this block seems no different with the next...

src/library/objects/ShowcaseExporter.js
```diff
@@ -0,0 +1,881 @@
...
+
+    ShowcaseExporter.prototype._splitText = function (text, ctx, maxWidth, splitter) {
+        if (typeof splitter === "undefined")
+            splitter = " ";
+
+        if (text.indexOf(splitter) === -1)
+            splitter = "";
+
+        var rows = [];
+        var words = text.split(splitter);
```
sinsinpub: split by space might only work for names in western lang

src/library/objects/ShowcaseExporter.js
DynamicSTOP: it was `max` before but i didn't liked how it looked on image

src/library/objects/ShowcaseExporter.js
```diff
@@ -0,0 +1,873 @@
...
+            img = this._statsImages[i];
+            drawStat.call(this, img);
+        }
+
+
+        for (i in this._statsImages) {
+            if (!this._statsImages.hasOwnProperty(i))
+                continue;
+
+            if (i === "or" && this.aircraftTypes.indexOf(KC3Master.slotitem(equip.masterId).api_type[3]) === -1)
```
sinsinpub: seems not work as `this.aircraftTypes` will become `[]`. just use `KC3GearManager.landBasedAircraftType3Ids` will do. `api_distance` is not suit for all carrier based aircraft: type 21/22 are not able to deploy to land base.

src/library/objects/ShowcaseExporter.js
```diff
@@ -0,0 +1,873 @@
...
+        var fontSize = 10;
+        ctx.font = fontSize + "pt \"Helvetica Neue\", Helvetica, Arial, sans-serif";
+        ctx.fillStyle = "#000";
+
+        var startX = x;
+        var img;
+        var noStats = true;
+
+        function drawStat(img) {
+            if (typeof equip[i] !== "undefined" && equip[i] > 0) {
```
sinsinpub: should be `equip[i] !== 0` because some equipment like those bulge hold negative stat.

src/library/objects/ShowcaseExporter.js
```diff
@@ -0,0 +1,873 @@
...
+
+        canvas.height = canvasData.height + this.rowParams.height * 2 + 18;
+        canvas.width = canvasData.width;
+
+        ctx.fillStyle = "#FFF";
+        ctx.fillRect(0, 0, canvas.width, canvas.height);
+        ctx.drawImage(canvasData, 0, this.rowParams.height * 2, canvasData.width, canvasData.height);
+
+        var created = "Made in KC3 on " + dateFormat("dd mmm yyyy");
+        ctx.font = "400 14pt \"Helvetica Neue\", Helvetica, Arial, sans-serif";
```
sinsinpub: as we discuss at discord, create a function to build these font string, and use `px` instead of `pt`.

Javran:
I just tried this branch, definitely a great feature. maybe in future when dealing with #1566, we can try to use steganography.js and encode some JSON data into it (just like what we have done with battle data)

so back to this feature itself, I have few suggestions about equipment list, better explained by picture:

2017-02-08_153308_3840x1080_scrot

could you make the number on top the total one ignoring improvements? this is to make it consistent with what we have in equipement list of the strategy room:

2017-02-08_153632_3840x1080_scrot

users would expect there's always a total number somewhere, instead of just those without improvements; and instead of say `+ 4star x1`, we just do `4star x1`

also if you look carefully at the `+ 10star x1` part of Daihatsu, you'll see that line is a bit cutted out, this problem may need some care.

I haven't took a look at the source code yet, as @sinsinpub has gave you some comments so just try first addressing that.

DynamicSTOP:
@Javran font cutting is done because i've used pt instead of px. Should be fixed in next commit.

Javran:
@DynamicSTOP I see. I forgot about ship list and has one more thing to add: I think one stat worth showing in the picture is luck: if one ship has more that usual luck (either through marriage or modded with maruyu) you might consider present that in addition to her level.

src/library/objects/ShowcaseExporter.js
```diff
@@ -0,0 +1,873 @@
...
+    ShowcaseExporter.prototype._drawShip = function (x, y, ship, background) {
+        this.ctx.fillStyle = background;
+        this.ctx.fillRect(x, y, this.rowParams.width, this.rowParams.height);
+        var xOffset = this.rowParams.height / 7;
+
+        this._drawStat(ship.fp, x + xOffset, y, this.rowParams.height / 2, this.rowParams.height / 2, "#ff8888", background);
+        this._drawStat(ship.tp, x + xOffset + this.rowParams.height / 2, y, this.rowParams.height / 2, this.rowParams.height / 2, "#00CCFF", background);
+        this._drawStat(ship.aa, x + xOffset + this.rowParams.height / 2, y + this.rowParams.height / 2, this.rowParams.height / 2, this.rowParams.height / 2, "#ff9900", background);
+        this._drawStat(ship.ar, x + xOffset, y + this.rowParams.height / 2, this.rowParams.height / 2, this.rowParams.height / 2, "#ffcc00", background);
+        this._drawStat(ship.lk, x + xOffset + this.rowParams.height, y, this.rowParams.height / 5, this.rowParams.height / 2, "#66FF66", background);
```
DynamicSTOP: @Javran luck is shown here if it's maxed out, though it might be good to show that luck was affected in any way.

Javran: @DynamicSTOP I see, maybe showing actual numbers might be more helpful, as I suspect not many players would have one or more luck-maxed out ships..

src/library/objects/ShowcaseExporter.js
```diff
@@ -0,0 +1,955 @@
...
+                if (response.data.UserRemaining > 10 && response.data.ClientRemaining > 100) {
+                    $.ajax({
+                        url: 'https://api.imgur.com/3/image',
+                        method: 'POST',
+                        headers: {
+                            Authorization: 'Client-ID 088cfe6034340b1',
+                            Accept: 'application/json'
+                        },
+                        data: {
+                            image: dataURL.substring(22),
```
dragonjet: you can use this instead:
https://github.com/KC3Kai/KC3Kai/pull/1803/files#diff-4834798524872cdd6bda1e77051ff285R185

DynamicSTOP: Currently it's PNG only, that's why i don't change it. My personal opinion is `jpeg is good only for photos, for everything else png should be used`.

dragonjet: ah I see, no problem then :+1:

src/library/objects/ShowcaseExporter.js
DynamicSTOP: @Javran I am not sure that they want see more numbers then it's already have. [From the example](http://i.imgur.com/XIGBXp8.png). But maybe we can add something like vertical progress bar that would represent luck going from min to max instead of just green stripe.

DynamicSTOP: From my experience developers and users are seeing usability by completely different angles.

Javran:
it all looks good to me. maybe just a minor issue: the ship exporting function ends up opening a blank tab if my output mode is set to "New Tab".

but I suspect this could be the case that URL is too long to handle for chrome. my `dataURL.length` is `2331858`

Javran:
now I can confirm this is [a chrome-side issue, as the URL limit is 2MB]( https://bugs.chromium.org/p/chromium/issues/detail?id=69227), which means `2 << 20 = 2097152` so mine exceeds the limit.

maybe we can also inform user about this in Q & A part, other than that, I think this is ready to land. good work!

Javran:
@DynamicSTOP I can confirm the fallback mechanism is working

dragonjet:
this would also close #1693

Javran:
I guess this is good enough for merging, but why this is still in WIP label?

PR #1793, opened 2017-02-07
Contributing Guidelines

### LINKED ISSUE #850: Convention for updating the translation submodule?
```
numachi% git diff
diff --git a/src/data/lang b/src/data/lang
index a94a4dc..b05f252 160000
--- a/src/data/lang
+++ b/src/data/lang
@@ -1 +1 @@
-Subproject commit a94a4dce2eefa08774f0b0f727a107e99898e629
+Subproject commit b05f2526e622dc9f9f7fb70d6952e708a5150800
```

When the submodule is updated, there is a change made in the containing repo that tracks where the current HEAD in the submodule is pointing.  Do we want to make commits regularly to update the main repo, or just on release, or for every commit on the translation repo (which seems like a waste)?

### LINKED ISSUE #1707: Setting guidelines
see https://help.github.com/articles/setting-guidelines-for-repository-contributors/ for detail.

we can put some conclusion that we have reached in #1698 there. just making a clear guideline so everyone can read and follow. especially PR-first and giving approvals - by doing that we can land a patch faster, as long as we have received few approvals from others, we don't have to wait @dragonjet for it. (for me it also save my time switching between local branches and cherry-picking patches everywhere.)

dragonjet:
* Close #1707 Guidelines about Pull requests
* Close #850 Convention for updating the translation submodule
* #1730 GitHub labels meanings

### DIFF
--- a/.github/CONTRIBUTING
+++ b/.github/CONTRIBUTING
@@ -1,0 +1,90 @@
+# KC3改 Contributing Guidelines
+
+### Contents
+* Getting Started
+  * Forking
+  * Cloning
+* Codebase
+* Incorporating your changes
+* Testers and Dev builds
+* Translators
+
+----
+## Getting Started
+### Forking
+* 
+* 
+
+### Cloning
+* 
+* 
+
+----
+## Codebase
+
+### Scripting Laguage
+* 
+* 
+
+### 
+
+----
+## Incorporating your changes
+
+### Making Pull Requests
+In a discussion in [this GitHub thread](https://github.com/KC3Kai/KC3Kai/issues/1698), it is now highly recommended to create features and enhancements on branches, and making pull requests to master.
+
+Here are few tips in making pull requests:
+* Include screenshots
+* State what the chages are / affected pages
+* Tag involved or relevant contributors
+* You can optionally make a to-do list if you created a pull request while it is a work in progress, so followers can see progress.
+* Tag the issue thread the update is fixing
+   * Optionally mention issue threads in your pull request description, and if done in correct format, the issue will be closed automatically when your changes are merged.
+
+The following example, when included in a pull request description will automatically close the specified thread, in this case [issue 1707](https://github.com/KC3Kai/KC3Kai/issues/1707), once the pull request is merged.
+```
+Fix #1707 Add file for contributing guidelines
+```
+Learn more about [comment tagging here](https://help.github.com/articles/closing-issues-via-commit-messages/). Although it says "_via commit messages_", this also works when included in pull request descriptions.
+
+### Approval of Pull Requests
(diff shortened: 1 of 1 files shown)


Javran:
refering to my own comment: https://github.com/KC3Kai/KC3Kai/issues/1895#issue-214534010
I feel we need to emphasize that untested or unconfirmed new features should never be pushed directly into master branch (but perhaps we can allow an urgent fix).

We need this for quality control, it's important that we have a working master branch with as less bug or controversal stuff as we can. now I think of this, it's not just for users but for contributors:

- as I stated previously, we should never put our user at risk of things like unconfirmed feature without proof, or any other untested features, ask our testers or other developers first. (btw if some users are willing to try our master branch, we can kindly invite them to our testing team, maybe put an invitation statement in README.md, so people interested can easily spot it. it's never a bad thing to have more active testers around)

- also the master branch would be a common start point for contributors that don't have direct access to this repo, so it's important to keep it stable so each individual don't have to do some "fixed this by the way" thing that leads to merge conflicts

I also think we need penalty for violations, maybe a removal of write permission, otherwise it's nothing and people don't listen or don't regret their misbehave.

ethrundr:
Regarding this issue, here's my opinion:

1. For those who directly push to master branch: unless it is an emergency fix that require the push to be immediately done to master. This should NOT be happened. As for the penalty, revoking their write permission temporarily might be a best option - as for how long, let community decide based on how severe the violation is.

2. If the same violation is re-occurring again and again, then, with all my respects, we can simply kick him out of the organization for good. We don't need any people like that, even if he contributes most to the team. Let him be an outside contributor if he wish to keep contributing, using a proper way of creating PRs before merges.

3. All PRs should contain complete list of what you did inside the code. Not all members of the team are understand codes, thus a short-but-detailed information regarding what are you going to add to the master is needed. This will also gives any of our user that come to stop by here know what we're currently doing, testing or developing. If anyone found smuggling any unannounced update inside his PR, we can simply treat it as point 1: it is basically the same like pushing directly to master, after all.

These points might be harsh, but I do think it is needed to maintain the extension is usable and available for everyone.

Nya-chan:
I agree with all points raised by ethrundr. For 3) it shouldn't be listing of all changes file by file (there's a tab for that after all), but again, it should list all **functionalities** that were added/changed/removed. And the count of them should be as low as possible, ideally a single functionality per PR/merge (this is not always possible, of course, for example when you hotfix a few issues at once - but in that case all fixes should be listed).

DynamicSTOP:
Would like to add couple words to above. According to #1730 `status:testing` is replaced by `status:review`. Thus being said i assume review is not only look through code but also **checkout and test**.

dragonjet:
close for now to declutter PR list. should re-open in case anyone gets the chance to update the readme file

PR #1799, opened 2017-02-08
Expedition Scorer Improvement (reattempted)

### LINKED ISSUE #792: Suggestions for expedition scorer

### LINKED ISSUE #1530: Improvement for Expedition Scorer

Javran:
See #1784. original PR is closed due to a messy commit log.

the following list is for auto-closing some issues:

- closes  #792
- closes #1530


Javran:
update: just rebased to target master, as `dev-sr-exped` has been merged.

other than that, this PR is ready and open for comments.

Javran:
kinda want to bump this and see if anyone can review this, because Winter is Coming.

Javran:
let's allow ~6 hours before merging. just to make sure we have some window before maintenance to double check.

PR #1800, opened 2017-02-08
Expedition Table Interface

dragonjet:
For #1772

### DIFF
--- a/src/pages/strategy/tabs/expedtable/expedtable.css
+++ b/src/pages/strategy/tabs/expedtable/expedtable.css
@@ -162,7 +162,7 @@
 	height: 100%;
 	float: left;
 	margin: 0px 3px 0px 3px;
-	font-size: 17px;
+	font-size: 14px;
 	padding: 5px 0px 5px 0px;
 	text-align: center;
 }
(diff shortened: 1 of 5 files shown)


Javran:
just tried, all looks good but few things:

- I'm not sure why should we use `.html` when we can just `.text("◄")` instead? actually `.data("order", _ )` is the exact thing I want to avoid, I think that means maintaining two states that do the same thing (see my comment in https://github.com/KC3Kai/KC3Kai/pull/1800/files#diff-c0090d90fddf0f1558c70e271fc3eb3eR844)
- would the use of `!important` considered bad practice? I thought I've carefully programmed CSS selectors to have the intented priority for each of them.

dragonjet:
* I switched it out with an html entity because when I pulled it into my local repo, it was an unknown character, and showed up as a square in the browser.
  * this also caused the use of .data(), because .html() does not return the html entity nor the special character but something with the \u notation or something similar.
* I also avoided the use of `!important` in the original dark.css, and those that needed override, I usually remove from original css file and put into the `legacy.css`. but when I saw it used afterwards, didn't check when those were added, so I just went with the flow.

Javran:
@dragonjet I see, I'm good with `.data` then. just trying to remove some `!important` if proved unnecessary, and this will be good enough.

Issue #1468, opened 2016-08-07
[suggestion] optional notes on settings


Javran:
some settings are just hard to explain in short words:

![2016-08-07_114732_3840x1080_scrot](https://cloud.githubusercontent.com/assets/1096354/17460353/ca791ea6-5c94-11e6-9eed-aeca90630dc7.jpg)

it's hard to understand what it does by just looking at this. there are other examples like "Salt Treatment" and "Troll Comments".

so I'm suggesting to leave some places for translators: it doesn't inherit from English, when translators feel the need for more detailed explanation in his particular language, he can opt in by adding a translation to `SettingsSomething_NOTES` (assuming this is a note for `SettingsSomething`) and that note would show up along with that setting/

dragonjet:
agreed :+1:

PR #1802, opened 2017-02-08
Provide an option of disabling special quest adjustment

Javran:
- added `ConfigManager.spCounterAdjust`, default to `true`.
- when set `false`, recover old behavior that doesn't try to deal with shared quest counters
- new term added: `SettingsSpecialCounterAdjust`

@KC3Kai/translators see https://github.com/KC3Kai/kc3-translations/pull/379/files for examples. sorry that we don't really have room to explain this option properly, until things like #1468 can be done. so try to make a concise description.

### DIFF
--- a/src/library/objects/Quest.js
+++ b/src/library/objects/Quest.js
@@ -14,7 +14,7 @@ Quest Type:
 */
 (function(){
 	"use strict";
-	
+
 	window.KC3Quest = function(){
 		this.id = 0;
 		this.type = 0;
@@ -22,7 +22,7 @@ Quest Type:
 		this.progress = 0;
 		this.tracking = false;
 	};
-	
+
 	/* DEFINE
 	Fill object with already-formatted quest data
 	------------------------------------------*/
@@ -33,14 +33,14 @@ Quest Type:
 		if (data.progress) {
 			this.progress = data.progress;
 		} else {
-			this.progress =  0;
+			this.progress =	 0;
 		}
 		if (!this.tracking) {
 			this.tracking = data.tracking;
 		}
 		this.attachMeta();
 	};
-	
+
 	/* DEFINE RAW
 	Fill object with quest data from Raw API response
 	------------------------------------------*/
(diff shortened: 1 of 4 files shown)


src/library/objects/Quest.js
```diff
@@ -96,20 +96,28 @@ Quest Type:
...
 	@param {number} amount - progress amount to be increased. Default: 1
 	@param isAdjustingCounter - if true, prevent recursively inc on specific quest itself
 	------------------------------------------*/
-	KC3Quest.prototype.increment = function(reqNum, amount, isAdjustingCounter){
+	KC3Quest.prototype.increment = function(reqNum=0, amount=1, isAdjustingCounter=false){
 		var self = this;
 		var isIncreased = false;
+		if (! ConfigManager.spCounterAdjust) {
+			if (this.tracking && this.isSelected()) {
+				if (this.tracking[reqNum][0] + amount <= this.tracking[reqNum][1]) {
```
sinsinpub: just wanna remind: this old logic is completely incorrect. eg: assume amount is 2, if tracking current value == max value - 1, current will never reach max value as `current + 2 > max`.
fortunately there is no any `increment` caller passing the `amount` argument before.
don't be angry over being a experimental animal, be pro lul

Javran: @sinsinpub I was angry because I am trying to be professional: you are letting a known buggy feature out and making **every user**, not just our devs and testers experimental animals.

I didn't really check how the impl works in detail, but all I know is it's been there for at least a year and no one has reported having a problem with it. so I just want the old trusty behavior back.

sinsinpub: don't forget to remind **every user** update their Chrome to 49 at least, otherwise the old trusty one will be betrayed

Javran: @sinsinpub I didn't see that line coming from someone that agreed to use es6 for this project?

by the way how buggy is es6? is that even worse than a buggy counter that needs to be fixed at least 3 times?

do you really know that a program first get parsed before ever executing? show me how to execute an ill-syntaxed source code, I'll be fascinated.

well now at least **everyone** has learned how a pro maintains open source projects and understands about programs. I'm sure to take a note for myself. thanks!

PR #1805, opened 2017-02-10
Make it possible to disable idle timer

Javran:
I didn't notice this previously, but now I feel the new idle timer is slightly on the annoying side: most of the time I just let the game open and all I need is just to have secretary ship in sight. and new idle timer ruins that by fading out the whole game screen. more importantly there is no way to officially cancel that - perhaps you can set a rather big number but that doesn't sound right to me.

that's all motivation for this PR, changes (pretty much copied from my commit log):

- ~~new setting box type: `number`~~ `small_text` turns out to be a better option
    - ~~most of the code copied from `long_text`~~
    - ~~no verification on field~~
    - value is parsed and written back to UI upon change 
        (btw the input value was never parsed in the original version so one could only hope the input (as a string) is coerced properly to a number)
- no idle timer on a neg/falsy input (off by def)
    - cancel idleTimer when time set to falsy / neg val
    - disabled by default for a less intrusive behavior
    - make it consistent that user always give time in seconds

I wish this can be merged before our winter update for the reason I mention in the very beginning, so sorry for the short reviewing time window.

(Edit: btw I feel `isInvalid` from `SettingsBox.js` is a bit unreadable as we can just simply passing objects as result values instead of doing bitwise tests and that kind of stuff, this could be a future task, but for the purpose of this PR I'm not going to change that)

### DIFF
--- a/src/library/injections/dmm_takeover.js
+++ b/src/library/injections/dmm_takeover.js
@@ -4,17 +4,17 @@
 	var master = {};
 	var meta = {};
 	var quests = {};
-	
+
 	window.DMMCustomizations = {
 		apply: function(response){
 			console.log('Applying DMM customizations...');
-			
+
 			config =  $.extend(true, ConfigManager, response.config);
 			window.ConfigManager = config;
 			master = $.extend(true, KC3Master, response.master);
 			meta = $.extend(true, KC3Meta, response.meta);
 			quests = $.extend(true, KC3QuestManager, response.quests);
-			
+
 			this.windowFocus();
 			this.attachHTML();
 			this.layout();
@@ -23,7 +23,7 @@
 			this.clearOverlayHotKey();
 			this.exitConfirmation();
 			this.screenshotHotkey();
-			
+
 			chrome.runtime.onMessage.addListener(this.subtitlesOverlay());
 			chrome.runtime.onMessage.addListener(this.clearOverlays());
 			chrome.runtime.onMessage.addListener(this.questOverlay());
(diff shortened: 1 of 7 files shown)


src/library/injections/dmm_takeover.js
```diff
@@ -567,13 +572,13 @@
...
 			// Hide on mouse move
 			$(".overlay_idle").on('click', function(){
 				hideIdleScreen = true;
 				$(this).hide();
 			});
-			
+
 			// Receives and remembers the time when a network request was last made
 			return function(request, sender, response){
 				if (!ConfigManager.alert_idle_start) return true;
```
dragonjet: ah but this line should have been the toggle, if user sets it to zero, idle timer will not run

sinsinpub:
not agree the 'new type number', just use `small_text` will do:
```json
				"type" : "small_text",
				"bound" : {
					"min" :  0, // or -1
					"type": "Number"
				},
```

dragonjet:
milliseconds wont fit on small text though, but if we just make small_text a bit wider.

sinsinpub:
@dragonjet just do it

dragonjet:
actually no need, it's autodetect if less than 1000 it will treat it as seconds instead of milliseconds

sinsinpub:
already add tooltips to settings, should be better with `small_text`
image

Javran:
@sinsinpub would you mind pushing the change to this branch, also feel free to revert c67a382 because that's not really needed in this case.

sinsinpub:
@Javran np, but to push to this branch, resolve conflict needed, give me more minutes

Javran:
added another commit just to move config check to proper location as @dragonjet pointed out (not really need `<=` though as the input verification will prevent it), now this should be all set.

PR #1816, opened 2017-02-12
attempt for a smarter auto adjust counter

Javran:
an attempt to solve the quest counter auto adjusting problem once for all:

- detect both falling-behind and overshotting problems and try to correct them
- prevent an known incompleted quest from completion
- some unit tests included

I admit the code is a bit complicated, but in order to have every case covered, I feel it has to be.
please follow comments in code to see if my reasoning is clear to you.

and in case you are too lazy to read the code, I have also written some unit tests to demonstrate the expected behavior of this auto quest counter adjustment.

despite the fact that this PR is ready for reviewing, let's take our time and don't rush for approvals, as quest tracking is one of those important features, I wish this one can be good enough to replace the old one.

after this we may remove the feature introduced by #1802, after all if the quest counter can correct itself in all cases, that feature won't be necessary.

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -300,11 +300,6 @@
 					"type": "Integer"
 				},
 				"options" : { "label" : "" }
-			},
-			{
-				"id": "spCounterAdjust",
-				"name": "SettingsSpecialCounterAdjust",
-				"type": "check"
 			}
 		]
 	},
--- a/src/library/managers/ConfigManager.js
+++ b/src/library/managers/ConfigManager.js
@@ -27,7 +27,6 @@ Retreives when needed to apply on components
 				devOnlyPages		: false,
 				forceDMMLogin		: false,
 				apiRecorder			: false,
-				spCounterAdjust     : true,
 				updateNotification	: 2,
 
 				showCatBombs		: true,
--- a/src/library/objects/Node.js
+++ b/src/library/objects/Node.js
@@ -1192,10 +1192,6 @@ Used by SortieManager
 							case 15:	// 15 = AP
 								console.log("You sunk a AP");
 								sunkApCnt += 1;
-								if (! ConfigManager.spCounterAdjust ) {
-									KC3QuestManager.get(218).increment();
-									KC3QuestManager.get(212).increment();
-								}
 								KC3QuestManager.get(213).increment();
 								KC3QuestManager.get(221).increment();
 								break;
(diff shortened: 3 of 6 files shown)


Javran:
@sinsinpub about `maxCount <= 1` problem, I'm doing it on purpose, at first `maxCount === 1` is indeed a separated case, but then I figure if we just let `maxCount === 1` go through these calculations the result will be the same so I removed that part to simplify the code.

Javran:
Update: might still need to exclude F6 or do something else, `1/3` is marked with `50%` marker, which in turn corrects the counter to `2/3`.

sinsinpub:
finally notice it...F6, F7 are both 1/3 = 50%, no 80%. however other quests with maxCount: 3 seem not (1/3 no mark, 2/3 = 80%). god know devs

Javran:
@sinsinpub yep, I guess making F6 and F7 special case probably is a must.

Javran:
update: new auto adjustment works fine for me, let's allow 10 days and see if this is still fine.

sinsinpub:
As this PR make counter possible to 'go back', i guess we could also handle special cases like F7, F8

For maxCount: 3 quests, no matter progres is 0.5 or 0.8, assume they are two types:

 * Regular type in-game: 1/3 no mark, 2/3 50%. Just adjust them as this PR do. Known quests: Bd4, Bd5, Bw10, Bm2, C2, D2
 * Special cases in-game: 1/3 50%, 2/3 50%. Known as: F7, F8. Don't touch if current counter is 1 or 2, just adjust to 2/3 if current is 0 or max(3). As if real counter is 1/3, only +1 and then go quest list in game, our counter will still go from 3/3 to 2/3.

Javran:
update: making `isCompleted` check earlier so that we have a chance to adjust multi-counter quests if they are completed

FYI for @KC3Kai/developers try avoid using `for ... of` syntax for now if you want to write some unit tests for it. we will need to include `babel-polyfill` for this syntax to work I've tried simply including that package and it didn't work.

Javran:
@sinsinpub I think what you are purposing worth a separated PR, as I feel this PR is a bit complicated already and I want to make sure it is correct before adding more new stuff. but you can build your own PR on top of this one, and make use of unit tests to demonstrate the intended behavior.

Javran:
also for @KC3Kai/developers: continued from my message above: if you can figure out how to make `babel-polyfill` work for unit tests that's more than welcome.

Javran:
update: removed toggle for special quest counter adjustment.
I've been keeping the adjustment feature on and experienced no problem. even if there are, this new auto adjust feature should correct them as soon as user visit their quest lists, so now we can safely remove feature introduced by #1802.

so now this PR is officially ready for merge, let's allow 8 hrs for questions and after which point anyone can feel free to merge this.

Javran:
merging without further ado.

PR #1827, opened 2017-02-14
Exporter: stripped ships from equip stat bonuses.

### LINKED ISSUE #1825: Possible bug with showcase export
In the new showcase export, the 4 corners seem to be colored if the corresponding stat is maxed - FP, Torp, Armor, and AA (clockwise from top left). However, my Maya K2 is colored:

image

Even though her AA is not maxed:
image
image

Maybe it's something to do with equipment? Or it's just me.
Chrome version: 56.0.2924.87 (Official Build) (64-bit)
KC3Kai version: 31.0.6

DynamicSTOP:
Close #1825

### DIFF
--- a/src/library/objects/ShowcaseExporter.js
+++ b/src/library/objects/ShowcaseExporter.js
@@ -73,8 +73,10 @@
             dv: "api_baku",
             ht: "api_houm",
             rn: "api_leng",
-            or: "api_distance"
+            or: "api_distance",
+            lk: "api_luck"
         };
+        this._modToParam = ["fp", "tp", "aa", "ar", "lk"];
         //TODO save next two vars somewhere
         this._outputMode = 2;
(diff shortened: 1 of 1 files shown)


sinsinpub:
how about just see if shipMaster stats[0] + shipObj.mod[fp/tp/aa/ar] reaches stats[1] like ship list does?

DynamicSTOP:
First func in sight was [this one](https://github.com/KC3Kai/KC3Kai/blob/master/src/pages/strategy/tabs/ships/ships.js#L777).
I see there are 5 ints in mode. ship.mod[4] is luck, right?

sinsinpub:
yep, 5th is luck. check ship master stats[0] + mod[?] == stats[1] should be the same way with game's, not need to care about equips and that function.

PR #1722, opened 2017-01-08
New Gameplay Mode

dragonjet:
A smaller pull request than #1645

* This branch now doesn't remove API Link and DMM Frame gameplay options
* Is only and additive to existing background script, doesn't replace existing functions
* Doesn't replace existing injections
* Only includes special DMM mode only, does not have online translations which the other branch already incorporated
* Doesn't remove many settings that the other branch deprecated

## Updates
| Area | Description | Screenshot |
|---|---|---|
| **Menu Updates**  | Smaller API link and DMM frame options  | 
| **New Play landing page** | Opens when selected the "_Play on DMM Website_" option  | chrome_2017-01-23_07-59-12 |
| **New idle Timer** | Shows up after configurable idle time. Slowly fades to darker screen for 100 seconds, Click to dismiss until next idle time  | chrome_2017-02-07_12-24-20 |

## Related Issues
* #1641 Chrome Version 55: Black screen
* #1643 [Issues] Panel don't start
* Close #425 Screenshot Feature Re-enabling In the Direct DMM Website Gameplay
* Close #1033 DMM网页模式 截图不能?
* Close #1274 Game freezing on right click in frame
* Close #890 OSAPI

### Customization
- [x] Customizable Background Images
- [x] Cuztomizable Game Frame Margins

### Overlays
- [x] Subtitles
- [x] Quest Translation Overlay
- [x] Non-TL'ed Quest GT link
- ~~Annoying Taiha Alert~~
- [x] Map Markers Overlay
- [x] Idle timer

### HotKeys
- [x] <kbd>F7</kbd> Toggle keyboard focus
- [x] <kbd>F9</kbd> Trigger screenshot
- [x] <kbd>F10</kbd> Clear all current overlays

### Other
- [x] Exit confirmation
- [x] Screenshot
- [x] Fit Screen

### Integrations
- [x] New menu option
- [x] New landing page for "Play" menu option

PR #1833, opened 2017-02-16
Patch  31.0.7

dragonjet:
### Updates
* Also show F33 tooltip with factors for Combined View
  * Adjust width of summary columns
* Compute enemy air power for researching unknown node
  * Key problem is to guess unknown slot capacity of abyssal ships
  * Hover on air battle condition to see known air power and unknown IDs
  * Should be fail-safe for unexpected input
* Show TP tooltip for all types of Combined Fleet
* Add double click link to ship drop
  * Fix double click data and event cleaning method
* #1825 Exporter: stripped ships from equip stat bonuses (PR #1827)

### Translations and Data
* Update WhoCallsTheFleet DB
* `en` quests
* `en` terms
* `jp` quests
* `th`quests

### Notes


puroidrogeno:
is quite long time that using DMM on website and API link stuck the game on a withe screen, from the today patch (the only one i can use) DMM on frame reach the game normally i can click everything and hear the sound but all the game is a black screen

dragonjet:
this is in the patch notes
* #1789 Release 31.0.0

in chrome 55+, dmm frame and api link may not work 

more info
* #1645 Merged Gameplay Option
* #1722 New Gameplay Mode

original issues posted during the first chrome mess-up
* #1641 Chrome Version 55: Black screen
* #1643 [Issues] Panel don't start

puroidrogeno:
@dragonjet i write because i think is a new problem, already checked those thread before write, this is how i have the game page using dmm website and api link: https://gyazo.com/b93c4a066eabf2a10f3189a5f4274881

instead using dmm on frame is like the normal game page with sound and clickable things but all is black, i mean i can play but at the same time i can't see what' appening

ch version 56

dragonjet:
@puroidrogeno can you try disabling kc3 (not uninstalling since it will clear your history) just uncheck it from extensions page and see if the same problem persists on dmm website

puroidrogeno:
@dragonjet hey buddy, i tried to clear as you asked, checked again flag/adblock and other stuff following the post on facebook, but still i got a withe screen when i get on kancolle page (i only can see dmm button on top, ads on right side and the glossary under the "game". i think is this version: 56.0.2924.87 causing problem, i just tried on other browser like firefox and the game start perfectly

dragonjet:
@puroidrogeno join us in discord. let us not discuss this in a patch notes thread.

PR #1835, opened 2017-02-16
Enhanced Handling of Gameplay Modes

dragonjet:
## Updated menu
* New Title for playing via the new game mode
* Removed DMM Frame and API Link from menu
* Less colors, inclines more to minimalist

## Updated Play page
* Toggle colors are now red and green
* Temporarily removed links in favor of **Game Modes** section
* Can select old gameplay modes here if user still wants them

chrome_2017-02-16_16-53-41


sinsinpub:
SS of JP landing page:
image

off-topics:
 1. too small line height for those smaller font-size lines may cause characters for other languages too compact, esp when chrome's min font size is 12px. (my ss is using 11px, so no problem)
 2. the height 100% not required to be 'fixed'.

sinsinpub:
problem found: ``, `` are still used in terms, but png file are removed

PR #1848, opened 2017-02-22
Update LBAS air power

sinsinpub:
Improve LBAS air power:

DONE

* Update default config of fighter power bonus data for land bomber: now known that they use dive bomber's bonus, not fighter's.
* Support latest calculation method for both regular fighter power and air defense interception power on enemy air raid
* Also support bonus of recon plane deployed to LBAS
* Update panel LBAS view

see [wikia](http://kancolle.wikia.com/wiki/Land-Base_Aerial_Support#Fighter_Power_Calculations)

TODO

 * As usual, we are not able to only update user's local config attribute of air power, have to tell ppl click 'Reset to default' button at settings page on release notes.

SS:
image


sinsinpub:
previously travis building passed but github failed to get its state for pr 😓

Issue #1832, opened 2017-02-16
[Suggestion] Migrate KCBadge to updated version


Nya-chan:
The current KCBadge has been abandoned by its author and the site for export isn't updated. The author published it on github, though, and the development continues.

I am now hosting the latest version at http://www.sanya.moe/kcbadge/ and plan to update it as long as the github is updated, so it should be possible to direct the export there.

ethrundr:
Is it possible if you create a pull request instead of making an issue?
It'd be a great help if possible to be done.

Nya-chan:
I'll look into it (haven't worked with GH that much so far, but it will be handy in the future, so why not) and try~

Javran:
let's keep this issue open for keeping track.

Vultren:
we can close this now. strategy room points to the right location,

PR #1862, opened 2017-02-28
badge generator location change

Vultren:
Following up on #1832. working with @Nya-chan to keep the badge generator updated.

### DIFF
--- a/src/pages/strategy/tabs/badge/badge.html
+++ b/src/pages/strategy/tabs/badge/badge.html
@@ -13,7 +13,7 @@
 <div class="page_padding">
 	<div class="page_section">Introduction</div>
 	<div class="section_body">
-		This page will export your data to <a href="http://threebards.com/kaini" target="_blank">Kai Ni Badge Generator</a> so you won't need to input the data yourself. We <a href="https://github.com/KC3Kai/KC3Kai/issues/1028" target="_blank">worked with the developers</a> of the tool for this integration.
+		This page will export your data to <a href="http://www.sanya.moe/kcbadge/" target="_blank">Kai Ni Badge Generator</a> so you won't need to input the data yourself. We <a href="https://github.com/KC3Kai/KC3Kai/issues/1028" target="_blank">worked with the developers</a> of the tool for this integration.
 	</div>
 	<div class="page_section">Data to include</div>
 	<div class="export_box section_body export_parts bscolor2">
@@ -115,7 +115,7 @@
 
 <div class="factory">
 	<div class="submit_form">
-		<form action="http://threebards.com/kaini/index.php" method="post" target="_blank">
+		<form action="http://www.sanya.moe/kcbadge/" method="post" target="_blank">
 		</form>
 	</div>
 </div>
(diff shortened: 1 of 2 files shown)


Javran:
just a reminder that http://www.sanya.moe/kcbadge now only supports English, if you switch to another language there'll be error message saying `Can't find DB, please contact Ryuuhou Kagiyama on Himeuta.net` that should be fixed before we can proceed to merge.

Vultren:
Wait, really? ok, let me check that out tomorrow, and if the problem is what I think it is, it should be fixed in the Zara Kai Ni Update

Nya-chan:
Just notifying that the last time I tried, the page couldn't get the post data, not sure the server has it sorted already and should be tested first.

Vultren:
there was a comma missing before CT that was causing the problem, I'll change the issue number in a few hours

dragonjet:
@Nya-chan @Vultren once this is ready, let us know so we can merge.

I'd also like to add you in the about page, as the original `ThreeBards.com` has an entry in the partners section. Would you like to be added separately as fork developers or under the partners tab as one entry for `sanya.moe`?

Nya-chan:
@dragonjet I have no problems with being mentioned there, you can go on with that~ Not sure what is the difference, so whichever is more fine with you, I guess?

Javran:
on an unrelated note, I wanna point out that I-26 is of the same class as I-19, so they are supposed to be categorized under same class.

Nya-chan:
Just a note that this should be ready to go shortly - it probably worked already, because I was stupid and was testing with sanya.moe instead of www.sanya.moe (like Vultren properly uses) and because of permanent redirect I wasn't getting proper POST data - one always learns that laziness doesn't pay off.

I'll just update the link to lead into github fork, since there's no point to direct people to report problems in a forum thread anymore.

Javran:
@Vultren @Nya-chan btw I feel something is slightly messed up across different languages:

- all Winter 2017 ship appear only in en
- Iku and Nimu are grouped together in non-ens but in en they are still separated

also some problem in Poster Collection related to remodel chain:

- Fujinami Kai doesn't seem to toggle Fujinami on (same problem goes with Asakaze Kai, Matsukaze Kai, Zara Due, I-13 Kai, I14 Kai)
- Saratoga icon not toggled on

Before testing this branch I've already remodelled all ships except Saratoga so not sure whether they have been supported in the first place.

Vultren:
@Javran is this all in the non-en versions?

Vultren:
ok I see what you mean with the toggle, that was me trying to get the update in before I went to work. that's an easy fix. Sara's in the conversion json though...... And I just now understood the "mstId2FleetIdTable" field in the conversion json. and Sara's entry is missing a one. the non-en on the other hand, that's going to take a bit of work. This was a problem before I took over the updates.

Javran:
those problems that I pointed out can have their own separated issue in k2badge repo. now that every language is working so I see no problem merging this.

I intended to open a new issue to put my comments there but got a bit confused: there are 2 forked repos from original author: https://github.com/Vultren/k2badge https://github.com/Nya-chan/k2badge but which one is supposed to be the main branch?

Vultren:
Stick the issues in mine. I'll be doing most of the updates

Javran:
since this PR has received some approvals I'm going to merge this. 
for problems that I've pointed out, they will be tracked by https://github.com/Vultren/k2badge/issues/2

Issue #1863, opened 2017-02-28
SR ship drop record


Javran:
the idea is to have a viewer to show ship drop statistics of different maps from users' sorite history.  see the very quick and dirty screenshot in https://github.com/KC3Kai/KC3Kai/issues/1858#issuecomment-282945354

I've pushed my code to `dev-shipdrop` but won't have time for further development, so feel free to take over.

for now there is no UI control. one has to change [these two lines](https://github.com/KC3Kai/KC3Kai/blob/6362d61b5e3cc82449257ed1dce53f4c62332a89/src/pages/strategy/tabs/shipdrop/shipdrop.js#L28-L29) in order to show drops from different maps.

and few of my notes:
- since db queries are naturally asynchronous, a worker should fit this job perfectly.
- note that two edges might end up in the same node, but I haven't dealt with this problem.
- UI controls:

    - a list for all regular maps and event maps
    - event map only: checkboxes for different difficulty
    - checkboxes for different ranks

107nationalstation:
Is the 2nd note you mentioned the meaning that , for example , the ship drop information that should be showed under node: B but mistake for node: A?

My English is poor , thanks.

Javran:
@107nationalstation that's not what I meant. to explain this you first need to understand that drop records are tagged by edges, not nodes. some edges might end up in the same node. for example in map 4-2 there are two edges `C -> D` and `H -> D` with both ends in `D`, but in my naive impl node D appears twice because it's edge-indexed (a correct impl should merge these two node D together)

107nationalstation:
@Javran I mistake what you mean because I always play kancolle in Chinese and I don't pass cet6.I get your point as I'll try to figure the code out at tomorrw class.

Javran:
@107nationalstation don't worry, pin me on twitter [at]JavranW if you wanna speak in Chinese xD

gensuihime:
Seems mostly merged and is now dead.

PR #1878, opened 2017-03-06
add event map at drop_record in SR and add ship link to ship info page

107nationalstation:
better UI at drop record page in SR


107nationalstation:
related to [issue 1863](https://github.com/KC3Kai/KC3Kai/issues/1863)

Javran:
sounds like you have more plan to do.
just for saving some more PRs, I'll just leave this one open for now, you can continue working on it and let me know when you are done.

107nationalstation:
@Javran My problem is that I lost all my data of event as a result It's hard to test the event map is cureently added.

Javran:
@107nationalstation don't worry, I can confirm event drops are working fine.
talking about UI itself, however, I'll prefer dropdown boxes rather that the current UI of selecting worlds & maps, it's taking too much space but not doing much: user will alway click ">" several times just to skip regular maps to be able to see event drops, in my opinion it will be much better to just show two dropdown boxes, one for world selection and another for map's. and this leaves more free space for more filtering features (e.g. event difficulty checkboxes, rank checkboxes) to be added.

107nationalstation:
@Javran Oh dropdown box is better choice thx for your guiding. the online backup feature is the one that I really want , but it's hard to use dropbox in some area. I'll test if it should work to put backup in my vps.

107nationalstation:
@Javran  now it looks like that

PR #1880, opened 2017-03-07
Filtering ships in ship list

Nya-chan:
### Content
- Filtering by string in a ship name
- Showing only dupe ships

### Limitations
- Doesn't do a system call, can probably be optimized
- Works with only hiding the filtered-out ships, which is okay except for when the pages are displayed, since they currently don't get recalculated

Screenshot


Javran:
I don't understand why this would be any different than a Ctrl+F in that page

Nya-chan:
@Javran Can't say for sure about the string filtering, but the dupe filtering is practical when you want to quickly compare multiple ships and find out which one is the "correct one" to scrap without cluttering around, for example. Though thinking about the same can be told about the Ctrl + F approach - the main advantage is no clutter from the other 190+ ships (in my case) preventing me to see whether I really found all I wanted with a string I used. (I hope that makes sense, I am sometimes bad with words w)

Nya-chan:
Can I poke this? It's been opened for some time already and I'd like to close it - whether it's accepted or not, that's up to the team, but it shouldn't hang around opened w

sinsinpub:
i just troll: in the past year, if i do not get in, 'hang arouond open' is frequently occurring thing 😂 

personally i don't hate the filter, but beside some codes improvement (will post by reviewing later), i also want to request more features:
 * Filter by the name of romaji. Considering current user is not in English, he type 'kuma' will match nothing, and have to input '隈' or something in his translation, which means not convenient at all, and no different with Ctrl+F. Always filter by reading of shipgirl (romaji, in fact the English ver) will be more useful.
 * Make the 'romaji input method' as a module, which could be reused at Compare page. yes, that page use such a filter method to find ship, and not convenient for non-English user.

src/pages/strategy/tabs/ships/ships.js
```diff
@@ -238,6 +238,54 @@
...
 				$(".ingame_page").hide();
 			});
+            
+            /* Filter for duplicates */
+            $(".dupe_filter").on("click", function() {
+                if (!$(this).hasClass("active")) {
+                    $(".ship_list .ship_item").each(function() {
+                        var thisShipId = parseInt($(this).find(".ship_id").text());
+                        /* Find master ID */
+                        var result = $.grep(KC3StrategyTabs.ships.definition.shipCache, function(e) {
```
sinsinpub: `KC3StrategyTabs.ships.definition.shipCache` => `self.shipCache` is okay

src/pages/strategy/tabs/ships/ships.js
```diff
@@ -238,6 +238,54 @@
...
+                            return e.id === thisShipId;
+                        });
+                        var dupeCount = 0;
+                        function getDupeShipsCount() {
+                            var dupeCheck = $.grep(KC3StrategyTabs.ships.definition.shipCache, function(e) {
```
sinsinpub: this part is really strange... `getDupeShipsCount` return nothing but set the `dupeCount `, declared unused `dupeCheck` var. if condition of dupe: true is matching the master id of base remodel, might use `RemodelDb.originOf` on two sides of `bid` (master id of ship)

sinsinpub:
@ethrundr hmm, seems: no. for `jp`, `tcn`, the English translation of ship name is completely not loaded. that's a problem need to be handled too 😂

ethrundr:
If this is going to be implemented, both version of the name should be loaded, despite one of them is not shown.

Not all people using JP and TCN are using the kanji to type and not all people who is not using JP and TCN are using romaji to type.

Nya-chan:
So basically - aside of fixes mentioned by sin_sin - we want to always have JPN + Selected TL languages in ships and search based on those? What about JPN, that one probably shouldn't load JPN + JPN, right?

Nya-chan:
(Also not sure we aren't overcomplicating it - if someone comes to the ship list, they see what language is used there and can search based on that. I think JPN users are used to kanji conversion throughout searches.)

ethrundr:
From my perspective, it should load by default the JP(Kanji) and Romaji, regardless of the language, since I believe no language employ a specific translation for ship names, right?

This way we can enable the search field to search both type of the name: Kanji or Romaji, without any additional settings/buttons.

ethrundr:
This indeed will be bit more complicated on our side, but it'll make the user more convenient, since the search bar is not language-specific.

Adrymne:
From personal experience, support for romaji -> kanji varies but hiragana -> kanji is common; e.g. searching くま (kuma) would match both 熊野 (Kumano) and 球磨 (Kuma).

Nya-chan:
@Adrymne Sadly that switching doesn't work for indexOf, apparently orz
@sinsinpub Could you check this? Added kanji and base (no prefixes / suffixes) kana for filtering, tested and it should work. Also Travis fails the build (jshint) on something that's not my code orz

sinsinpub:
already checked and refine things, see branch `dev-shipname-filter`. will use it if u don't mind my changes
(impossible to merge this pr until u fix the duplicated `execute` function 😓

Nya-chan:
OK, no problem, will close this, then~

PR #1883, opened 2017-03-08
Fix time unit of API handling and refresh panel materials in time

sinsinpub:
As there are many codes changed, open a PR for more tests to check if I missed something or not.

~~No feature changed~~, just unify and refine time handling codes when timestamp converted to seconds or hours, and add time unit to vars name. 

At first, my motivation is the fix thingy. After double checking of each API handling, I have to clarify and explain the enhancement of feature: Keep tracking with game displaying for 8 basic materials (fuel, ammo, steel, bauxite, torch, bucket, devmat, screw), show their latest values in time and correctly at devtools panel.

*Long story*: as we know, there are 3 types of API call which would consume or gain materials:

 1. In api_data of response, there is `api_materials` liked data to tell us all the absoluate values of materials the player hold.
we have used these values as `resources` cached in `Player.lastMaterial` array and record them into IndexDB every hours.
API examples of this type: `api_port/port`, `api_get_member/material`, `api_req_hokyu/charge`, ...
 2. There is no materials data in API request or response, but we (KC client side) are able to infer the increasement or decresement values from other API attributes or known shipgirl data.
we have recorded these changes into IndexDB for ledger, but do nothing with the `Player.lastMaterial` cache array.
API examples: `api_req_kaisou/remodeling`, `api_req_kousyou/createship`, `api_req_quest/clearitemget`, `api_req_nyukyo/start`, `api_req_mission/result`, `api_req_air_corps/set_plane`, ...
 3. No materials data in API and only KC server side update the exact values of resources.
we can only try to compute them by ourselves.
This type of API is mainly new game mechanism about LBAS and jet bombers such as: `api_req_map/start_air_base`

This PR also done:
Check all the API handlers, of type 2 above, add `resources` and `consumables` updating and refresh event triggerring for panel, aside the ledger db recording codes.
This is why I said 'need more tests', to check if all the corresponding APIs are handled and are the values in panel always the same with in-game's.

Tech details:

 * Two frequently used functions at global.js: `toUTCseconds`, `toUTChours`
 * Replace all `Math.hrdInt("floor"...` converters with previous ones
 * Add time unit to those `ctime` vars, such as `...Seconds` or `...Hours`, fix those function calls with wrong time unit argument.
 * Let more API calls update Resources & Consumables in time (might via delta values only)
 * Improve PlayerManager:
   * Let all side-effect only functions return `this` instance for chain usage
   * Clearify acceptable time unit of parameters for `setResources` and other time related functions
   * Let `setResources`, `setConsumables` support delta values updating

### DIFF
--- a/src/assets/js/global.js
+++ b/src/assets/js/global.js
@@ -372,6 +372,12 @@ String.prototype.hashCode = function() {
 			return ret;
 		}
 	};
+	
+	Number.prototype.valueBetween = function(lfs, rfs) {
+		lfs = lfs === undefined ? -Infinity : lfs;
+		rfs = rfs === undefined ? Infinity : rfs;
+		return Math.min(Math.max(lfs, rfs), Math.max(Math.min(lfs, rfs), this));
+	};
 }).call(Number);
 
 /* JS NATIVE CLASS */
--- a/src/assets/js/global.js
+++ b/src/assets/js/global.js
@@ -161,7 +161,13 @@ Date.safeToUtcTime = function(date) {
  * Convert String to UTC timestamp/1000.
  */
 Date.toUTCseconds = function(dateStr) {
-	return Math.floor(Date.safeToUtcTime(dateStr)/1000);
+	return Math.hrdInt("floor", Date.safeToUtcTime(dateStr), 3, 1);
+};
+/**
+ * Convert String to UTC timestamp/1000/3600.
+ */
+Date.toUTChours = function(dateStr) {
+	return Math.hrdInt("floor", Date.safeToUtcTime(dateStr) / 3.6, 6, 1);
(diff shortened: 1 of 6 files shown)


src/assets/js/global.js
```diff
@@ -366,6 +372,12 @@ String.prototype.hashCode = function() {
 			return ret;
 		}
 	};
+	
+	Number.prototype.valueBetween = function(lfs, rfs) {
```
Javran: might just do `function (lfs=-Infinity, rfs=Infinity)` here?

sinsinpub: that also works. i just intendedly keep this file with no ES6 syntax for now... 😅

Issue #635, opened 2015-08-14
About Event Sortie Page


ReiFan49:
now after map difficulties is totally supported,
i guess there's something that need to be checked back about event sortie page...
1. is it intentional that non-cleared map use no gauge instead `curhp` and `maxhp` ?
   Status: **OK**
2. for every sortie, it might have a difficulty data attached to it (new db ver right ...). what i'm thinking is, decorate the info box either in total collapsed form or at least expanded (not fully expanded is OK, but not collapsed) with difficulty marker.
   
   what i'm thinking is similar like this
   ![Easy](https://cloud.githubusercontent.com/assets/9847780/9287425/ab431da4-433e-11e5-824f-cf8f2db6bdc5.png)
   ![Normal](https://cloud.githubusercontent.com/assets/9847780/9287396/c9761bf2-433c-11e5-9274-6ab11fa5baf8.png)
   _explanation: a text with "my usual rotation" style, attached with a gradient that makes the sortie_box more colorful (but there's **no rainbows** for any difficulty marker, easy green, normal orange, hard red)_
   Status: **OK**
3. decorate the map clear (not per sortie, but the map selection) with difficulty marker (either a medal like in-game, or something that i cannot think off...)
   Status: Ongoing (using single-letter marker)

dragonjet:
for event maps, non-cleared has `...hp`, but for ExOps, it has `kills`

I agree with 2, where they can also see which difficulty they got a drop or something.

and 3 is something important too, on which difficulty it was cleared.

btw I forgot game mechanics but can one change difficulty after clearing?

ethrundr:
my own opinion, since I loved something simple and not too flashy:

for no.2 and 3, if it's using just a simple difficulty marker, I'm on for it. but if it's involving gradient or such, somehow I'm objecting it. the current one is already good and simple enough, no need to make it more flashy.

if its purpose to notify only the difficulties, only marker is enough in my opinion.

however, at any rate, this will be important to be implemented. just keep it as simple as possible, yet still informative.

dragonjet:
ah for 2 and 3 we'll just add the ribbon image beside the map name

ReiFan49:
ok, i'll wait the event badge to come,
while waiting, the number 1, is... lol
"Not cleared" + `noclearnogauge` class on event maps.
that's why i'm asking this xD (gonna push this one)

ethrundr:
[...]

Issue #56, opened 2015-05-08
View total sorties to a map


URSvAir14:
I was wondering whether it is possible to count the total number of sorties made to a map. Hopefully this will include both event maps and standard maps.

PR #1885, opened 2017-03-09
Total sorties count on maps, medals in event lists

Nya-chan:
## Features
- Show count of total sorties in all normal worlds/event or world/map by returning the total of the sorties KC3Kai registers before breaking it down into pages. As requested in #56.
Screenshot
- Show medals on cleared event boxes for easier visual orientation (tried to avoid making them too flashy). Pointer events disabled as to prevent obscuring box functionality, CSS only, so can be turned off in styles. Added the medals to assets, since the project didn't have them so far.
Had the idea with #635 which can now probably be closed.
Screenshot


Nya-chan:
Update screenshot:
Screenshot

Vultren:
Maybe a bit bigger on the text size. Like "total sorties on top and the number below it?

PR #1876, opened 2017-03-06
Add external backup to Dropbox service

reinforce:
Add ability to upload/download kc3data files directly to/from Dropbox storage.
This can be useful for external backup and sync via different devices.

Basically this is a simple Dropbox client which uses REST API.
You need Access Token to use this feature. See the help section to learn how to get it.

For now sync can be done semi-automatically:
 * Export data to file in "Local Backup"
 * Select file for uploading in "External Backup"

And it would be uploaded to Dropbox
And back:

 * Download file from list
 * Import data in "Local Backup"

Related to #17
screenshot at 2017-03-14 13 33 13

Signed-off-by: Aleksei Mamlin <mamlinav@gmail.com>

PR #1890, opened 2017-03-13
Upgrade jQuery version to latest 2.x

sinsinpub:
Minor upgrade, from 2.1.3 to 2.2.4, no migrating guide found. Suppose to be compatible. For #1876 

Just to test: grunt built package works or not, because version removed from file name. (not sure is version in file name really necessary? @dragonjet)


reinforce:
@sinsinpub I think it's better to keep version in file name for fast version checking.

sinsinpub:
@reinforce i don't feel necessary... for human, check it at first line comment. for computer, access `jQuery.fn.jquery` will be fast enough.

Javran:
actually jquery is the only thing that has a version in its name, now that the version number is in head comments so we didn't loss any info.

dragonjet:
yep, jquery version in filename is good :+1:

Issue #1891, opened 2017-03-13
Hitting tab on levelling page messes up textboxes


Accelyte:
### Versions:
Chrome Version 56.0.2924.87 (64-bit)
Windows 10 64-bit
KC3 v.31.1.1 Webstore using recommended playing method

### Steps to reproduce:
Hit tab on the levelling page. Easy as that.

### Actual behaviour:
![chrome_2017-03-13_23-00-06](https://cloud.githubusercontent.com/assets/21363469/23878811/7bc0654a-0841-11e7-8925-debf8b220b04.png)

### Expected behaviour:
I dunno...you tell me. All I know is that something's broken.

Nya-chan:
This happens since the tab "jumps" into the textboxes which are actually always there, just shifted below the text.

A solution might be to give every input tabindex of -1, so they are not selectable via tab at all, if that is reasonable (and if we want them to be selectable when they appear, we can always edit this value via jQuery).

I'll have a look at it.

PR #1893, opened 2017-03-14
Tab fixes leveling

Nya-chan:
Straight fix for #1891 - disallowing any tabbing in leveling room via tabindex = "-1" in selects and inputs except for the moment when the fields are edited (in case somebody uses tab in that moment).

Fixed the visual for the input with level number while I was at it, so now it shouldn't cut off awkwardly.


sinsinpub:
i think there is simpler way: add `display: none` to `ship_input` class. as hidden input will not get focus by tab key 😂  no js modification needed.

keep the input control styles fix and also fix the select control

sinsinpub:
should fix that by commit b0bb6eb. will close this, still thanks for ur efforts.

Issue #1678, opened 2016-12-21
[Bug] Akashi repair timer reset after retrieving from fleet record


HKGPsy:
Since the introduction of fleet record, we can put Akashi repair fleet into a record and allow it to repair without occupying a fleet. Within that 20 minutes, we can use any ships to go for sortie, even those who are being repaired by Akashi. The repair timer would not reset as long as we retrieve the Akashi repair fleet from record before 20 minutes repair time is up. This is an extremely powerful feature as we can simultaneously repair ships and going sorties. However, in KC3Kai, the Akashi repair timer does not take into account this function and would reset when we retrieve it from the fleet record.

I would suggest fixing this bug and also include a Akashi timer on the panel to countdown for the remaining repair time (even when Akashi is not currently the Flagship) so that we know when to retrieve the repair fleet from the record.

Please let me know if you need more explanation on the details, thanks.

nonsensed:
This is a bug in the game and in theory is bannable. I don't know if this is something that should be supported by a viewer. The devs for kancolle have already proven they are willing to ban users for bug exploitation of various kinds.

Side note: Akashi currently doesn't work correctly in kc3 last I checked - #1378

HKGPsy:
It is unclear if its a bug or intentional as a function of fleet record. Indeed, there are no official announcement of such a features so it being a bug cannot be ruled out. My guess is its going to be a grey area that would never be fixed (just like team 2 flagship never sink), but I cannot be sure and I agree it might not worth it to take the risk.

ps. Currently the logbook [航海日誌](http://nekopanda.blog.jp/) supported this function, so if we decided not to include this function in KC3Kai, we can still complement it with the logbook, just fyi.

PR #1896, opened 2017-03-16
Fix #1378 Rewrite Akashi timer to match ingame behaviour

### LINKED ISSUE #1378: Wrong Akashi Timer in the main panel
In the current setting, Akashi timer in the main panel would reset every time you go back and forth to the lobby. However, this is not correct since the timer won't reset if you don't change the fleet. This bug makes the Akashi timer totally useless since it will always give you wrong count on the time.

Adrymne:
The repair timer should behave as follows: a repair (re)starts when the
player modifies their fleet from the Organize screen, and the resultant
fleet has meets the correct conditions (Akashi flagship, not red/orange
HP, not in docks). Modifying the fleet in other ways - i.e. scrapping,
modernization, equipment change - will not affect the timer.

When the player visits the homeport and at least 20 minutes have passed
since the repair started, HP will be recovered and the timer resets. If
it has been less than 20 minutes, no action is taken.

The number of ships affected by the repair is based on the number of
Ship Repair Facilities equipped by Akashi at the time the repair is
applied.

(This is my first time contributing to a public repository, so hopefully I didn't make too much of a mess of things...)

### DIFF
--- a/src/library/managers/PlayerManager.js
+++ b/src/library/managers/PlayerManager.js
@@ -46,6 +46,7 @@ Does not include Ships and Gears which are managed by other Managers
 				new KC3LandBase(),
 				new KC3LandBase()
 			];
+			this.akashiRepair = new KC3AkashiRepair();
 			return this;
 		},
 
@@ -348,7 +349,7 @@ Does not include Ships and Gears which are managed by other Managers
 				return this;
 			}
 
-			this.fleets.forEach(function(fleet){ fleet.checkAkashi(); });
+			this.akashiRepair.onPort(this.fleets);
 
 			var
 				// get current player regen cap
--- a/src/library/modules/Kcsapi.js
+++ b/src/library/modules/Kcsapi.js
@@ -432,6 +432,7 @@ Previously known as "Reactor"
 		// Use a Preset
 		"api_req_hensei/preset_select":function(params, response, headers){
 			var deckId = parseInt(params.api_deck_id, 10);
+			PlayerManager.akashiRepair.onPresetSelect(PlayerManager.fleets[deckId-1]);
 			PlayerManager.fleets[deckId-1].update( response.api_data );
 			PlayerManager.saveFleets();
 			KC3Network.trigger("Fleet", { switchTo: deckId });
@@ -668,14 +669,16 @@ Previously known as "Reactor"
 					}
 					// If not the same fleet, also recheck akashi repair of source fleet
 					if(oldFleet !== fleetIndex-1){
-						PlayerManager.fleets[oldFleet].checkAkashi(true);
+						PlayerManager.akashiRepair.onChange(PlayerManager.fleets[oldFleet]);
+						PlayerManager.fleets[oldFleet].updateAkashiRepairDisplay();
 					}
 				}
 			} else { // Remove ship
 				PlayerManager.fleets[fleetIndex-1].ships.splice(changedIndex, 1);
 				PlayerManager.fleets[fleetIndex-1].ships.push(-1);
 			}
-			PlayerManager.fleets[fleetIndex-1].checkAkashi(true);
+			PlayerManager.akashiRepair.onChange(PlayerManager.fleets[fleetIndex-1]);
+			PlayerManager.fleets[fleetIndex-1].updateAkashiRepairDisplay();
 			PlayerManager.saveFleets();
 			KC3Network.trigger("Fleet", { switchTo: fleetIndex });
 		},
--- a/src/library/objects/AkashiRepair.js
+++ b/src/library/objects/AkashiRepair.js
@@ -1,0 +1,93 @@
+/* AkashiRepair.js
+
+Manages the timer for a player's Akashi repairs.
+*/
+(function () {
+  "use strict";
+
+  var MS_PER_SECOND = 1000;
+  var MS_PER_MINUTE = 60 * MS_PER_SECOND;
+
+  window.KC3AkashiRepair = function () {
+    this.timer = new KC3AkashiRepair.Timer();
+  };
+
+  /*--------------------------------------------------------*/
+  /*----------------------[ GETTERS ]-----------------------*/
+  /*--------------------------------------------------------*/
+
+  KC3AkashiRepair.prototype.isRunning = function () {
+    return this.timer.isRunning();
+  };
+  KC3AkashiRepair.prototype.canDoRepair = function () {
+    return this.timer.getElapsed().canDoRepair();
+  };
+  KC3AkashiRepair.prototype.getElapsed = function () {
+    return this.timer.getElapsed().ms();
+  };
+
+  /*--------------------------------------------------------*/
+  /*---------------------[ LISTENERS ]----------------------*/
+  /*--------------------------------------------------------*/
+
+  // Listener for api_req_hensei/change
+  KC3AkashiRepair.prototype.onChange = function (fleet) {
+    if (KC3AkashiRepair.hasRepairFlagship(fleet)) {
+      this.timer.start();
+    }
+  };
+
+  // Listener for api_req_hensei/preset_select
+  KC3AkashiRepair.prototype.onPresetSelect = function (fleet) {
+    if (KC3AkashiRepair.hasRepairFlagship(fleet) && !this.timer.isRunning()) {
+      this.timer.start();
+    }
+  };
+
+  // Listener for api_port/port
+  KC3AkashiRepair.prototype.onPort = function (fleets) {
+    if (this.timer.getElapsed().canDoRepair()) {
+      var akashiFlagExists = fleets.some(KC3AkashiRepair.hasRepairFlagship);
+      if (akashiFlagExists) {
+        this.timer.start();
+      } else {
+        this.timer.stop();
+      }
+    }
+  };
+
+  /*--------------------------------------------------------*/
+  /*------------------[ INTERNAL CLASSES ]------------------*/
+  /*--------------------------------------------------------*/
+
+  var Timer = function () { this.startTime = undefined; };
+  Timer.prototype.start = function () { this.startTime = Date.now(); };
+  Timer.prototype.stop = function () { this.startTime = undefined; };
+  Timer.prototype.isRunning = function () { return !!this.startTime; };
+  Timer.prototype.getElapsed = function () {
+    return new KC3AkashiRepair.DeltaTime(this.startTime);
+  };
+  KC3AkashiRepair.Timer = Timer;
+
+  var DeltaTime = function (startTime) {
+    this.startTime = startTime;
+    this.now = Date.now();
+  };
+  DeltaTime.prototype.canDoRepair = function () {
+    return this.ms() >= 20 * MS_PER_MINUTE;
+  };
+  DeltaTime.prototype.ms = function () {
+    return this.now - this.startTime || 0;
+  };
+  KC3AkashiRepair.DeltaTime = DeltaTime;
+
+  /*--------------------------------------------------------*/
+  /*----------------------[ HELPERS ]-----------------------*/
+  /*--------------------------------------------------------*/
+
+  // Returns true if the fleet's flagship is of the Repair Ship class, false otherwise
+  KC3AkashiRepair.hasRepairFlagship = function (fleet) {
+    var flagship = fleet.ship(0);
+    return flagship.master().api_stype === 19;
+  };
+})();
--- a/src/library/objects/AkashiRepair.js
+++ b/src/library/objects/AkashiRepair.js
@@ -16,6 +16,20 @@ Manages the timer for a player's Akashi repairs.
   /*----------------------[ GETTERS ]-----------------------*/
   /*--------------------------------------------------------*/
 
+  // Calculate the amount of HP that can be repaired,
+  // and the amount of time until the next point of HP can be repaired
+  KC3AkashiRepair.prototype.getProgress = function (dockTime, hpLost) {
+    if (!dockTime || !hpLost) { return { repairedHp: 0, timeToNextRepair: 0 }; }
+    var elapsed = this.timer.getElapsed();
+
+    if (!elapsed.canDoRepair()) {
+      return KC3AkashiRepair.calculatePreRepairProgress(elapsed);
+    }
+
+    var tickLength = KC3AkashiRepair.calculateTickLength(dockTime, hpLost);
+    return KC3AkashiRepair.calculateProgress(elapsed, tickLength, hpLost);
+  };
+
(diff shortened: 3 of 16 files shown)


Javran:
Thanks! the code is LGTM, also like the fact that this is coming with some unit tests xD.
I'll have a try on my local branch before approval.

Javran:
could be a rare case, but I'm getting something funny, not sure how it happened though as I wasn't really paying attention while akashi is repairing:
 2017-03-17_124049_3360x1080_scrot

the time gets stuck at that value until fleet compo is changed.

Adrymne:
I found the issue - `initAkashiRepair()` wasn't being called when you loaded a fleet preset. More generally, the interaction between presets and the Akashi repair is not handled at all.

Should it be? The general impression I got was that it was a bug and potentially bannable (see [wikia](http://kancolle.wikia.com/wiki/Akashi#Anchorage_Repair_.28.E6.B3.8A.E5.9C.B0.E4.BF.AE.E7.90.86.29), #1678). Even if not, the output should probably be something more sensible - I'm open to suggestions.

Javran:
my suggestion is not to think too much about that bug, just treat it as if player is manually configuring a fleet. trying to deal with bugs not coming for our side is just asking for trouble. plus, if I understand this mechanism correctly, it'll be back on track anyway after next repair gets applied. that's already way better than before.

we can worry about whether to exploit that bug later, maybe providing user an option to activate it and tell them to use it at their own risks, but I think that can totally be a separated story.

sinsinpub:
i'm wondering if the mechanisms / facts are really understood by all of us... would u please list up the differences between old implementation and yours? more words, we should list up all known facts about Akashi facilities, and recheck things, like which ones our old codes not do but your new codes do, which ones old codes are wrong but your new codes correct them.

besides the preset fleet installation resetting timer issue mentioned above, known facts u did not metion yet: for example, how about fleet sent to exped? another example, how about multi flagship Akashi in two or more fleets?

Adrymne:
Javran's [comment](https://github.com/KC3Kai/KC3Kai/issues/1378#issuecomment-230084883) from the referenced issue covers the original motivation for the PR - after applying a repair, the timer for the next repair was not being properly reset to 20 minutes. It also fixes a case where the timer was resetting when it shouldn't have been (scrapping/modernization).

However, in the process of putting together the summary of the Akashi repair mechanic, I found at least one bug and some cases that need further testing. I don't have time to finish it now - hopefully it'll be done sometime tomorrow - but it seems there is still work that needs doing.

Javran:
@Adrymne I'm good with a placeholder instead of showing an inaccurate number, but please avoid using `--:--:--` because it's the same as ships that does not require any repairing and this could confuse users:

2017-03-19_030334_3840x1080_scrot

also, perhaps you are already doing this: refresh the timer when there are repairs detected, no matter of whether it is a regular one or preset one. I think there are users that exploiting this bug so they don't really want to fiddle with the fleet to get the timer back.

Adrymne:
### Timer (Re)start
* Timer starts when Akashi is set as flagship of a fleet and she is not at red/orange HP or in the dock
  * Adding/removing/reordering ships of that fleet from the Organize (編成) screen resets the timer
* The following do not reset the timer: inspecting (詳細) a ship, modifying other fleets, equipment change, docking a ship, doing PvP, scrapping a ship, using a ship as modernization fodder
  * *Does docking Akashi cancel an in-progress repair?*
* (!) If Akashi was in the docks, she will start repairing when the player visits the home port
  * *Does visiting the dock screen work as well?*
* Morale/fatigue does not affect repairs
* (!) Sending the fleet to an expedition resets the timer
  * (!) Timer pauses while the fleet is on expedition
  * Timer does not pause during a sortie
    * *Does sortieing reset the timer?*
* Multiple fleets can be repaired at once with multiple Akashis
  * Actions that trigger resets only affect 1 fleet (2 if the player is moving ships between fleets)
  * (!) *Simultaneous repairs share a common timer?*
    * From wikiwiki: 当然複数艦隊で旗艦を明石にしても機能するが、**修理時間のタイマーは共通となっている**ため注意が必要
    * Can't test this one since I only have 1 Akashi
* (!) [Presets](https://github.com/KC3Kai/KC3Kai/issues/1678) can be used to perform repairs
  * The exact mechanics are unclear and it could be considered a bug

### Repair targets
* Repairs 2 + number of "Ship Repair Facility" equipped ships
  * Targets based on fleet position, ineligible targets are not skipped
  * It counts the number of cranes when the repair is applied
* Ships at red/orange HP or in docks are not repaired

### Repair Application
* If at least 20 minutes have passed, visiting the home port will apply a repair
  * This is the only way to apply a repair; other ways of resetting the timer will not do so
* After a repair is applied, the timer is reset to 0
* Time needed to repair HP is calculated similarly to [dock repairs](http://en.kancollewiki.net/wiki/Docking#Repair_time)
  * (!) Time elapsed is rounded down to the nearest minute
  * (!) Total repair time rounds up to the nearest minute instead of adding 30 seconds
  * Will always repair at least 1 HP, even if it would normally take longer

### Repair Cost
* (!) Similar to [dock repairs](http://en.kancollewiki.net/wiki/Docking#Repair_cost), but rounds up instead of down
  * (!) Doing a full repair will use the same formula as the docks
* (!) Akashi will not repair a ship if the player does not have enough resources to fully repair that ship

Sourced from [enkcwiki](http://en.kancollewiki.net/wiki/Akashi#Anchorage_Repair_.28.E6.B3.8A.E5.9C.B0.E4.BF.AE.E7.90.86.29), [wikiwiki](http://wikiwiki.jp/kancolle/?%CC%C0%C0%D0#hakuchi) and my own observations. The items marked with (!) are not currently supported by the PR. I believe only the first three are regressions. Most are uncommon edge cases, but the repair prediction definitely needs fixing.

Adrymne:
This is a bit awkward, but I'm currently working on a rewrite of the PR; I did some further research and it turns out the repair works a bit differently than we thought.  The new implementation is primarily based on [notes](https://github.com/andanteyk/ElectronicObserver/blob/d9753ae917ed96f3fc3afd73a2b174e7bf0751b1/ElectronicObserver/Other/Information/kcmemo.md#%E6%B3%8A%E5%9C%B0%E4%BF%AE%E7%90%86%E3%81%AE%E4%BB%95%E6%A7%98) from the developer of Electronic Observer - which, based on my own testing, appear to be accurate (translation attached below, a check would be appreciated).

Which brings up a couple of questions:

- Should the rewrite be its own PR? A lot of the old code is either gone or moved to a new context (AkashiRepair is in its own file now, for one) and the functionality is expanded a fair bit; it's now arguably beyond the original scope of this PR (which was poorly named in hindsight). 
- How should the repair's interaction with presets be handled? I now know how to model it correctly and am inclined to do so, but as previously mentioned, at least some of the English community considers it a potentially bannable bug. The Japanese community [doesn't](http://ch.nicovideo.jp/nila/blomaga/ar904333) [seem](http://ch.nicovideo.jp/rorio/blomaga/ar904485) [to](http://ch.nicovideo.jp/rorio/blomaga/ar904095) (it's implemented in EO), though it isn't documented on wikiwiki either.

---

#### Anchorage Repair Specification

Definition of terms:

* Each player's account has a single, global `anchorageRepairTimer`, which stores the start time of a repair.
  * Its initial value is `null`. This value indicates that there is no ongoing repair.
* `start()` sets the current time to `anchorageRepairTimer`.  
`stop()` sets `anchorageRepairTimer` to `null`.
* A "repairable ship" is a ship that is not in the repair docks, not on expedition, and has HP > 50% and HP < 100%.

`start()` is called if, after performing a regular fleet composition action, the flagship is a Repair Ship. The action does not have to involve the Repair Ship. For example, it would be called in the following situations:

* A Repair Ship is set as the flagship
* A Repair Ship was second as the second ship, and the first ship is removed.
* A Repair Ship is the flagship, and the second and third ships positions were swapped.

'Composition' actions are those that use `api_req_hensei/change` (adding, removing, switching, drag+drop, remove all but flagship); **saving/deploying presets, scrapping and modernization do not count**.  
*(TL Note: Based on my own testing, the 'remove all but flagship' button does not count as a composition action, despite using the same API endpoint.)*
(Unconfirmed, but a sinking a ship is probably treated the same as scrapping).  
Performing a composition action on a fleet that does not have a Repair Ship as flagship does not affect the timer.  
When `start()` is called, **the condition of the Repair Ship (in docks, < 50% HP, etc) and the presence of Repairable Ships is not checked**. It only checks that the flagship is a Repair Ship.

Even if a composition action is performed mid-repair, `stop()` is not called (though it will call `start()` again).  
Because of this, it's possible to modify a fleet with presets without causing the timer to (re)start.

Also, it is confirmed that deploying a preset with a Repair Ship flagship will call `start()` if `anchorageRepairTimer === null`.  
(It's called at the moment that the preset is deployed, not when visiting the home port.)  
If `anchorageRepairTimer !== null`, deploying a preset will not cause `start()` to be called. 

The following checks are performed upon visiting the home port:

* If `anchorageRepairTimer === null` or it has been less than 20 minutes since `start()` was called, then nothing happens.
* If `Date.now() - anchorageRepairTimer >= 20 minutes`, then `canDoRepair()` is called:
  * If it returns `true`, HP is restored based on the elapsed time, and `start()` is called.
  * If it returns `false`:
    * If there is a fleet with a Repair Ship flagship, call `start()`
    * Else call `stop()`

`canDoRepair()` returns `true` if the following conditions are fulfilled:

* The flagship is not at orange HP
* The flagship is not in the docks
* The fleet is not on expedition
* There is at least 1 Repairable Ship at (0-based) position `n`, for which `n < repairRange` is true (where `repairRange` is equal to the number of Ship Repair Facilities equipped + 2)
* There are enough resources to perform the repair*.

(*) Natural resource regen is applied before checking. Thus, even if there aren't enough resources when the repair begins, it's possible to perform the repair if enough are generated in the meantime.

Experiment log:

```
05:28 Repair begins - HP54
05:48 Repair confirmed - within ~15 seconds, Inazuma is added (timer reset)
05:54 Inazuma scrapped
06:08 Repair confirmed - HP56
06:13 Yellow HP Akashi saved to preset, place different (unharmed) Akashi as flagship and begin repair
06:19 Preset with yellow HP Akashi deployed, put in repair dock, then unharmed Akashi immediately redeployed
06:33 Repair confirmed - HP57
06:42 Undamaged Musashi added - no Repairable Ships in range
06:45 Preset with green HP Ise deployed
07:02 Repair confirmed - HP58

09:49 Set chuuha Akashi as flagship (timer reset) - HP 30
09:52 Load unharmed Akashi from preset
10:09 Repair confirmed - HP 31
10:11 Set docked (chuuha) Akashi as flagship (timer reset) 
10:15 Bucket chuuha Akashi
10:31 Repair confirmed - HP32

10:56 After having a non-Akashi flag for 20+ minutes, deploy preset with Akashi (don't return to the home port yet)
11:17 Go to home port, repair confirmed - HP 33
11:38 Reset the timer in the same way (deploy Akashi from preset without going to homeport)
11:52 Go to homeport - no repair
11:58 Repair confirmed - HP34

14:20 Deploy Maya (timer reset)
14:29 Maya consumed by modernization
14:42 Repair confirmed

14:54 Deploy only full HP ships
15:15 Return to home port
15:25 Deploy preset with damaged ships
15:35 Repair confirmed HP68

15:47 Akashi as flagship, Amatsukaze (36/39 HP) as 2nd ship, send on Expedition 2
16:12 Return to homeport - no repair
16:16 Expedition returns - no repair
16:32 Repair confirmed - HP 37
```

For viewer implementation:

* `api_req_hensei/change`: if Repair Ship flagship, `start()`
* `api_req_hensei/preset_select`: if Repair Ship flagship and timer is `null`, `start()`
* `api_port/port`: if `Now - timer >= 20min`:
  * if Repair Ship flagship, `start()`
  * else `stop()`

This should more or less replicate Akashi's repair.

Reference: [[Preset Akashi Repair] A Hypothesis about the Akashi repair timer (JP)](http://ch.nicovideo.jp/nila/blomaga/ar904333)

Adrymne:
Apologies for the mess - I'm still fumbling my way towards proficiency with git (also I'm an idiot).

Am I okay to rebase away the extra commits? Should I disappear the first version as well?

richarddoge8079:
Replying to your question on Discord (sorry I was busy doing my research report), any changes to your fork wouldn't effect the main fork of the repo, so yeah go crazy on your fork of the repo. Just remember to test it out so when you do a PR there wouldn't be much trouble intergrating it to the `master` branch.

ethrundr:
As long as a PR hasn't been merged, no changes will be done in the main repo.

So feel free to wreak havoc on your own fork and your own copy.
When you feel all is done perfectly and passed the checks, then notify here so we can do the review and merge.

Adrymne:
Right, good to know.

There'll be one more commit to unify the logic for calculating the length of an Akashi repair in one place, then it'll be done unless someone finds something that needs to be changed.

Adrymne:
Okay, I'm done making changes unless it turns out there's something that needs fixing.

Adrymne:
Noticed a bug - the new `.calculateRepairTime()` can return results < 20 minutes. This is intentional, but something callers need to account for. I'll fix it and rebase.

Also, I know I said I wasn't going to add anything more, but the thought occurred last night: since this PR adds support for using presets in combination with Akashi repairs, should it also add a warning of some kind? Perhaps a popup when loading a preset with an Akashi flagship, which can be disabled from the Settings.

While I feel it's unlikely anyone would stumble on this by mistake - or get banned for using it, for that matter - it might be best to be safe.

ethrundr:
For now, I don't think it is necessary. I believe if someone set up a preset with Akashi as flagship, he has some purpose for it that is might be unknown to us.

sinsinpub:
will take more time to read codes and test ingame

ethrundr:
I'll merge this tomorrow so it can be tested, if no further notes are being added to this PR.

sinsinpub:
@Adrymne  a question: should `stop()` be called on a `composition action` only if flagship of all fleets are not Repair Ship?
image
because if we have to wait for 20 mins to `stop()`, this panel value will be keeping counting and show 'Repairing' even there is none Repair Ship  😓

Adrymne:
As per the notes from EO, the only time `stop()` gets called is when you visit the homeport after 20+ minutes and no fleet has an Akashi flagship; otherwise, it will continue to run.

(Granted, the only way to take advantage of this behaviour is through loading a preset, since composition actions that put Akashi as flagship will reset the timer.)

If you wanted to hide this fact, you could modify [this](https://github.com/KC3Kai/KC3Kai/blob/f04cae94bfa3b0fa5e60272f25b10cabf3a70545/src/library/objects/Fleet.js#L579) line to check whether Akashi is the flagship of the fleet before subtracting.

PR #1900, opened 2017-03-18
address #1895

### LINKED ISSUE #1895: What is this "Health Metric" in SR profile page?
I noticed this feature with very misleading description has sneaked into master branch:

> Not confirmed yet, but some proofs show that server counts them for you. And higher values you got, higher risk you would receive 201 catbomb.

Where is this "some proofs"? This is very wrong, otherwise every top 500 players should have been receiving 201 catbomb all the time.

I'll not repeat myself: **features with unconfirmed info should never go straight into master branch**. And I see no reason not to put this into our repo's guileline: **never put our user at risk of things like unconfirmed feature without proof, or any other untested features**.

From now on, I have zero tolerance of things like this. As long as I have the write permission to this repo, except a silently added & unconfirmed feature, or a buggy/untested one to be reverted at any time on master branch. And I encourage every developer to do the same - I'd rather having a commit war between developers than having a problematic release that upsets users.

Javran:
Fix #1895

- comment out "Health Metric" section in FAQ because it's simply wrong or lack of an actual proof or at least some sorts of evidence. And because the section name "Sortie Statistics" (see bullet right below) pretty much explains itself we don't need a FAQ section for it.

- rename "Health Metric" to "Sortie Statistics", we should not put value to users and "Sortie Statistics" is a neutral name that I think most of us would agree on.

### DIFF
--- a/src/pages/strategy/tabs/profile/profile.html
+++ b/src/pages/strategy/tabs/profile/profile.html
@@ -23,8 +23,12 @@
 	<div class="help_q">How to make my changed in-game "Profile Data" affect pages of this tab?</div>
 	<div class="help_a">Not necessary to use "Reload Button" of your Chrome, just click the "Heart Logo" on the top left side to reload latest data at any time.</div>
 	
+	<!--
+		this section is commented out because the statement
+		about having a higher sortie count being 201 catbomb is wrong or lacking evidence.
 	<div class="help_q">What does "Health Metric" mean?</div>
 	<div class="help_a">It shows how hard you have play this game (sortieing to normal maps) recently. Stay healthy gaming :) <br/>Not confirmed yet, but some proofs show that server counts them for you. And higher values you got, higher risk you would receive 201 catbomb.</div>
+	-->
 	
 	<div class="help_q">Can I export my whole history (not just profile data) and move it to another device?</div>
 	<div class="help_a">Go to "Data Backup" page.</div>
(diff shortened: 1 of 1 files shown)


Tedster59:
@dragonjet not exactly sure on this, thinking we may want to approve this also

dragonjet:
I didn't reply on the issue thread but I agree on this change :+1:

DynamicSTOP:
Never read or heard  about "health metrics" on both wikis or reddit. Can mislead people to think that it's related to ship health somehow. At least that's what i thought in first place. Like not getting taiha during sortie or something.

As for catbombs, i would prefer to see those proofs as a link in faq or remove that section. Never seen anything like sortie count in kc-bot community (for anyone who reads this, **bots are bad**, mkay? **don't use bots**, mkay?) and they recommend to just stop doing anything, not just sorties. I would rather think that it's api request count what gets you catbombed and show that number instead. Well i was bombed while upgrading my ships like 2-3 times in a row back in 2016, so i think sortie count isn't the main reason.

If sortie count is confirmed to be the reason of catbombs i would prefer to highlight those count rows with red\orange colours and add a tooltip for them explaining that you are above safe level and can get catbombed.

p.s. i shouldn't post links to bot discussions right? :smile:

ethrundr:
Never see terms nor proof about that.

True, some people, including me, also suspects that the number of repeated plays, or any others, might contributing to the 201 catbombs. However, there is no solid proof for such issue. All we had is just speculation and make-right calculations: just like how we deal with most mechanics in game - we don't know the exact calculation, but we made one ourselves in order to be able to closely reproduce what we had in game.

So, agree with the changes.

ethrundr:
Merging PR to master due to enough review gathered already.

PR #1913, opened 2017-03-23
Battlecount position hotfix

Nya-chan:
A fix for the sortie count that - for a reason unknown to me - got to release absolutely positioned and with no spacing around it. Fixed the positioning, otherwise left it untouched.

Should look like this (note the spacing):
Screenshot


sinsinpub:
it's not a fix, it's a fall behind.
also my fault, i was thinking there is enough space for the text on right. but i realize it's no.
when page count is bigger than 99, there will be not enough space for both total page and sorties count. with float right, the element will be wrapped to next line. like this:
image
so i use  a workaround to fix the position of text and keep it small enough for large number.

real fix should be considering following:
 * should not be wrapped no matter how long are page number and total numbers
 * other language text length and font size should be taken account if someday it's l10n.
 * default minimum font size of chrome should be taken account

Nya-chan:
> should not be wrapped no matter how long are page number and total numbers
Since we can get into thousands, that probably means own line? I can just put it above the aging and to the left in that case.

sinsinpub:
own line takes too many space of height, might be the final method.
i was also thinking take two small line (inside the 1 pagination line) for page count and sortie count on the right. but still need to be fixed 😂

Nya-chan:
> i was also thinking take two small line (inside the 1 pagination line) for page count and sortie count on the right.

That wouldn't help either once the paging got into thousands instead of hundreds, I think. To be honest I am not sure whether paging itself will stay on a single line once we hit those amounts w

Screenshot relevant:
Example

ethrundr:
I forgot whether my records is in hundreds or thousands already.
I'll get in touch in a few hours to see my own record, and if it's thousands, I'll show you how it's looked like on my side.

All in all, I think I need to update my latest checkout first 😂

sinsinpub:
as there is a line for batch switcher now... maybe just put the counts in the white space of that line 🤔

Nya-chan:
That's a good idea - I'll move it there

PR #1920, opened 2017-03-24
Show HP prediction of enemies in mouse-over

Tibo442:
Respects battle info config.

Example Overkill Example 
PvP Example Overkill Example


Tedster59:
definitely looks like a useful feature :+1:

sinsinpub:
iirc, someone also requested show the hp loss values of our fleet (not only by hp bar) before battle result screen

PR #1926, opened 2017-03-25
Add custom css setting for strategy room and panel box

Tibo442:
Full customisation of panel and strat room. Allow users to add own CSS on top of a theme in settings. 

Example panel
[Strategy room example settings](http://i.imgur.com/mvEDFAr.png) / [Strategy room example](http://i.imgur.com/zptTp76.png)

Proposed TL diff, English isn't my native language so I'll let someone else do translation terms. Feel free to change titles or description.
```diff
diff --git a/data/en/terms.json b/data/en/terms.json
index 2156248..f0f7f44 100644
--- a/data/en/terms.json
+++ b/data/en/terms.json
@@ -338,6 +338,7 @@
 	"SettingsBGSizeDesc" : "See CSS background-size property",
 	"SettingsBGPosition" : "Background Position",
 	"SettingsBGPositionDesc" : "See CSS background-position property",
+	"SettingsCustomCSSDesc" : "Close and reopen panel box or reload strategy room page to see changes. Don't forget that sometimes '!important' is needed to overwrite defaults.",
 	"SettingsPlayerScale" : "Game Player Scale",
 	"SettingsSubtitles" : "On-screen Subtitles",
 	"SettingsSubtitleFont": "Subtitle Font Family",
@@ -372,6 +373,7 @@
 	"SettingsPanelSize" : "Panel Size",
 	"SettingsPanelOpacity" : "Interface Opacity",
 	"SettingsPanelBoxBackColor" : "Panel Box Back Color",
+	"SettingsPanelBoxCustomCSS" : "Custom Panel Box CSS",
 	
 	"SettingsChuuiWarningDangerousFeature" : "Are you sure that you really want to change this?\nAny risks that you're attempting is beyond KC3改's responsibility.\n\nMake sure you're know what to do before.",
 	"SettingsErrorNG" : "Saved!",
@@ -384,6 +386,7 @@
 	
 	"SettingsStrategyRoom" : "Strategy Room",
 	"SettingsStrategyTheme" : "Strategy Room Theme",
+	"SettingsStrategyCustomCSS" : "Custom Strategy Room CSS",
 	
 	"InstalledTitle": "KC3改 has been installed successfully!",
 	"InstalledText": "Thank you for installing KC3改! We hope that these tools can help you play KanColle and make it a better experience! Do not hesitate to approach us if you need help. Check the about page to know where we can be reached.",
```

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -888,6 +888,15 @@
 					"placeholder" : "SettingsBGColorDesc",
 					"tooltip" : "SettingsBGColorDesc"
 				}
+			},
+			{
+				"id" : "pan_custom_css",
+				"name" : "SettingsPanelBoxCustomCSS",
+				"type" : "textarea",
+				"options" : {
+					"placeholder" : "SettingsCustomCSSDesc",
+					"tooltip" : "SettingsCustomCSSDesc"
+				}
 			}
 		]
 	},
@@ -904,6 +913,15 @@
 						[ "legacy", "Traditional" ],
 						[ "dark", "Dark" ]
 					]
+				}
+			},
+			{
+				"id" : "sr_custom_css",
+				"name" : "SettingsStrategyCustomCSS",
+				"type" : "textarea",
+				"options" : {
+					"placeholder" : "SettingsCustomCSSDesc",
+					"tooltip" : "SettingsCustomCSSDesc"
 				}
 			}
 		]
--- a/src/library/managers/ConfigManager.js
+++ b/src/library/managers/ConfigManager.js
@@ -189,8 +189,10 @@ Retreives when needed to apply on components
 				pan_bg_position		: "top center",
 				pan_opacity 		: 100,
 				pan_box_bcolor 		: "rgba(100, 100, 100, 0.618)",
+				pan_custom_css		: "",
 
-				sr_theme			: "legacy"
+				sr_theme			: "legacy",
+				sr_custom_css		: ""
 			};
 		},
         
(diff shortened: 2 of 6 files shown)


Nya-chan:
Agree with @ethrundr - since it injects code into head pretty directly, can't there be issues with XSS (like someone distributing a code to users which they won't check and will send their game IDs to the author?) or something along the lines?

Tibo442:
I tried a few methods to try some xss but failed, we could throw in some warning via the TLs just to be safe or a popup (like with In-page Refresh) when they edit it for the first time.

ethrundr:
I'm still opt for limited editable power to the user though.

Instead giving them full access to the script like that, better if we provide them a simple input field and let the script modify on their own, outside of the user direct involvement. This way it's more safer since we can limit the text length up to what is actually needed by the field.

dragonjet:
it's just style injection though, they cannot execute javascript that way. even so, anyone can edit any webscript just by going to DevTools > Sources > edit the file > Ctrl+S. tbh we don't really need CSRF protection since everything is client-side. He edits something, he breaks his own browser, game, and account. We do not have server-side handling that will make his code appear on other users. Also, MIT.

Tibo442:
For strategy room: the dark theme is ~37k chars long, legacy is ~27k
For devtools: natsuiro is ~64k and plain is ~21k

Made limit 64k should be enough in case a user makes a complete makeover of the theme. Pages still load fine after when using dark theme (37k chars) while having legacy selected. (Using, worst case, 128k out of 5m available in localstorage is hopefully fine). If someone really wants more they can manually edit config from console.

ethrundr:
If no one against the idea, or want to give other comments/remarks/revision, we'll clear to merge this I think?

PR #1933, opened 2017-03-26
Change shift-click replay export to right click + help + pvp

Tibo442:
- Change shift-click to right click
- Add information about shift-clicking ability in help section
- Add rightclick exporting to PvP

### DIFF
--- a/src/pages/strategy/tabs/pvp/pvp.js
+++ b/src/pages/strategy/tabs/pvp/pvp.js
@@ -36,8 +36,9 @@
 			});
 			
 			// Download replay button
-			$("#pvp_list").on("click", ".pvp_dl", function(){
-				self.downloadReplay($(this).data("id"));
+			$("#pvp_list").on("click contextmenu", ".pvp_dl", function(e){
+				e.preventDefault();
+				self.downloadReplay($(this).data("id"), e);
 			});
 			
 			// Show ship info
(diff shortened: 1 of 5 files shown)


src/pages/strategy/tabs/pvp/pvp.js
```diff
@@ -331,6 +339,13 @@
 						mapnum: 0,
 					};
 					
+					if(e.shiftKey) {
+						window.open("https://kc3kai.github.io/kancolle-replay/battleplayer.html#" + encodeURIComponent(JSON.stringify(encodeData), "_blank"));
```
DynamicSTOP: Looks cool. The only thing i want to mention though. Can it update buttons before opening new page?
Because otherwise it stays in "to site" mode. Not sure if it was part of the plan.

ethrundr:
Is it possible to add our own menu using right-click as trigger?
In that way, touch-screens counterpart will require to do a long-tap instead of normal-tap (click).

Tibo442:
Changed it to use right clicks instead of shift clicks, also added help to pvp.

sinsinpub:
a sad report: long tap will just trigger text selecting, not contextmenu 😂  well, nevermind, will make another support for touch event

ethrundr:
Merging this for now then, since there's no issue left to be addressed

PR #1939, opened 2017-03-28
replaced  DataURI with Blob in showcase exporter

DynamicSTOP:
Made based on [this so answer](http://stackoverflow.com/a/12300351). Special thanks goes to @Adrymne

1.  It allows to ignore 2 * 1024 * 1024 url size limit for new tab
2.  Browser window resize works WAY faster if Blob used in url instead of dataURI

### DIFF
--- a/src/library/objects/ShowcaseExporter.js
+++ b/src/library/objects/ShowcaseExporter.js
@@ -351,12 +351,16 @@
     };
 
     ShowcaseExporter.prototype._openInNewTab = function (dataURL, topLine) {
-        if (dataURL.length >= 2 * 1024 * 1024) {
-            this._download(dataURL, topLine);
-        } else {
-            window.open(dataURL, "_blank");
-            this.complete({});
-        }
+        var byteString = atob(dataURL.split(',')[1]);
+        var mimeString = dataURL.split(',')[0].split(':')[1].split(';')[0];
+        var ab = new ArrayBuffer(byteString.length);
+        var ia = new Uint8Array(ab);
+        for (var i = 0; i < byteString.length; i++)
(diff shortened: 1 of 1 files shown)


Adrymne:
Curious why you didn't use `canvas.toBlob()` - is Chrome 50 too recent?

On a related note, we should probably de-dupe the image export logic.

sinsinpub:
@Adrymne yes, a little. current kc3kai min chrome is supposed to be 49
@DynamicSTOP troll: why use a `*` in title, there is no markdown parsing 😂  will look if there is another method to get blob later.

DynamicSTOP:
@sinsinpub title was taken from first commit message and as for commit messages it's just a habit from project at work.

https://gist.github.com/DynamicSTOP/53c51e12dd56c9d257cef2344f9e2e40
Can't force imgur to use blob. Always error 500, but 500 is server side :\

DynamicSTOP:
Btw where are info about min chrome version hiding? Can't find anything in about\readme.md\manifest.json\webstore page

sinsinpub:
@DynamicSTOP no hard definition at anywhere (#1789 mentioned it), just because some known players still using kc3kai under Windows XP. dj was thinking that set min chrome version for webstore, but

Adrymne:
*Really* should have posted this earlier, but: the contents of this PR will be covered by the fix for #1942.

DynamicSTOP:
I am ok with that. The only thing that worries me is imgur upload. Can't force it to use blob. Keep saying error 500, but it should be something server side.

Adrymne:
I'm having issues with that as well; Imgur was apparently unable to read the blob. In the worst case, we can use the dataUrl as a fallback.

PR #1938, opened 2017-03-28
Revamped HP counter look

Nya-chan:
Resulting look:
Screenshot

Puts the diff into the slightly larger bar and shows predicted HP instead of current (while in prediction state). Doesn't change CF look (like previous commits targeting this section).

Targets issue #1934

PR #1941, opened 2017-03-29
Update cumulative

### LINKED ISSUE #1932: Event difficulty not hiding after closing fleet/nodes
Event difficulty not hiding after opening and closing fleet/nodes buttons.

[Example](https://streamable.com/ben7h)
Image

sinsinpub:
see #1938
close #1932 

sortie count position:
image
pvp
image


Nya-chan:
@sinsinpub (and maybe others?) - can I just ask if you fix such stuff that you assign yourself to the issue so I can see that I shouldn't do them? I assigned myself to these fixes, have them on the branch (was going to add some more and push today) and now the work can be pmuch thrown out. Not that I care about that fact "too" much - you did it better code-wise and added few more code fixes I would've skipped that improve the legibility - but it's still a free time loss for me.

ethrundr:
I think it's a common sense that left everything someone has assign themselves until the certain person is giving up for the issue.

It's kinda rude to take over someone's part when they've assigned themselves to the issue.

sinsinpub:
i don't think there is someone's part taken over. just do a quick fix for those things already get common view.
even some assigned issues had been solved by others, i don't think there will be time wasted. get in the same part of codes could learn from other's both good and bad. if the person don't know what to do and what has been done of others' part, how could make peer code review possible?
another bad example, if someone assign himself and never give up for a long time, what should happen then... ofc i'm urging nobody 😂

Nya-chan:
"another bad example, if someone assign himself and never give up for a long time, what should happen then"

Ask them in the issue, like we do with closing issues that seem dead.
I personally check notifications daily and if I were to leave for a long time or permanently, I would unassign myself from all the issues - that's a common courtesy for a project management.

"even some assigned issues had been solved by others, i don't think there will be time wasted"

Sorry, but it's literally time wasted for me. I know I can gain a bit from that - code-wise, that is - but that isn't true 100 % of the time and to be honest, if I could consciously choose between doing work that won't get used (just to improve slightly in code ways) and spending time with my fiancée, I'd pick the latter. I don't have as much free time as I used to have during high school / university and I value it more these days.

ethrundr:
> another bad example, if someone assign himself and never give up for a long time, what should happen then

simple, leave a reply, asking whether xxx is still working on it or not.
wait for a day or two for replies. if by then no reply was given, then you're clear to take over: we can simply deemed as xxx has not keeping up to date with the issues. make sure you leave another reply stating you're taking it over, though (or simply edit the last reply is enough - the point is the previous assignee knows that the job is being taken over)

> even some assigned issues had been solved by others, i don't think there will be time wasted

I've experienced this in a real-life working environment. While it's true if you seeing it from the whole project's progress, it is not when you're talking about the individuals inside. Taking over someone's job, especially if goes without any notice, might building up a serious frictions, especially when collaborating online. You may never know how far someone has progressed before they upload their work. Imagine if your work is only one semicolon short when you realized someone has taking it over without notice, and completing the work - rendering yours is useless. The point is only one: communicate your stance, what you want to do, to the others. Do not do it out of the blue, especially when others has picked it up first.

> if the person don't know what to do and what has been done of others' part, how could make peer code review possible?

You can always ask about the progress, can't you?
A simple "how's this issue is going?" is suffice. Also, I believe most of the developers joined up here is not that prideful. I've seen so many cases when we discussed about how and why stuff made, both here in GitHub, as well as in Discord (or Gitter, way before Discord).

All in all, all that we need is a simple communication. As mundane as it may looks, communicate between project members could prevent needless frictions - just like this one.

ethrundr:
Merging these cumulative updates as month has changed and tomorrow is maintenance anyway.

IMO if we're going to make cumulative updates for minor updates like this, we can release them monthly: at the 2nd or 3rd each month, we merged those minor updates into the master, so next updated version will have them.

Issue #1942, opened 2017-03-30
De-duplicate Image Output


Adrymne:
Showcase Export, Screenshots and Replay Image output should share the code for outputting an image.

sinsinpub:
definitely what i also want 👍 
remind: Uploading to imgur should be an independent, reusable module. Don't forget that images (urls) uploaded should be managed (saving to db, viewing at SRoom, etc)

Adrymne:
I'm not sure Imgur upload needs to be its own module? I was planning on having the default behaviour being what I described in #1936 (use user-selected method, then fallback if they fail), but `saveTab()`, `saveDownload()` and `saveImgur()` will also be exposed so they can be used directly if necessary.

Also, I was considering opening the imgur link as a new tab after the upload finishes. Thoughts?

sinsinpub:
i personally never use this feature to upload to imgur, because these uploaded anonymous things will not be controllable by myself 😂
there are two points:
 * imgur image link may be forgotten if only open in new tab once and never saved manually
 * not able to remove from imgur, because the management id (they call it 'deletehash') is not saved, even not shown.

although there might be few privacy info in these images, embedded replay data include more sensltive info
similar discussion: https://github.com/ShareX/ShareX/issues/389

Adrymne:
It'd be easy enough to exclude specified methods from the fallback. We may even want to expose that option to the end user.

Adding the ability to delete images is possibly a good idea, but probably deserves its own issue.

PR #1948, opened 2017-04-02
Deduplicate Image Export

Adrymne:
Refactor image output logic from `Screenshot`, `ShowcaseExporter` and `sortielogs` into new module `ImageExport`.

---

Mentioned this in #1942, but I suspect it might be a good idea to modify Settings so the player can choose to exclude certain output methods (or maybe just remove `saveImgur` as a fallback method). I could use some help with the UI design though; experience has taught me that particular area is not my forte.


DynamicSTOP:
This PR make me think that i have to start learning ECMA6 :rofl: i an not skilled enough in it to make a review.

Adrymne:
I can dial back the ES6 if it's an issue, most of it is just sugar. I do need Promises though - my first attempt using callbacks got messy in a hurry. As it is, the code should be fairly robust, even if it is kind of ugly.

(Ironically, I spotted a bug regarding ajax error handling as I was typing this.)

I could probably do better with async/await or co.js, but the former is too recent (Chome 55) and I don't know what the policy on including new libraries is for the latter.

DynamicSTOP:
Nah, es6 features should stay. Supporting es6 was discussed here #1698  and if i'm not mistaken everyone was ok about it. It's just i who need to learn it :rofl:

Adrymne:
I'll probably do some rebasing tomorrow. The end result shouldn't change, but b319b71 should really be three separate commits: "De-duplicate to new module" + "Implement output fallback" + "Prevent new tab from grabbing focus".

Adrymne:
This ended up being a bit more substantial than the quick rebase I expected. Changelog:

- Fix Image Quality setting not being properly applied
- Fix unable to download image from "New Tab" mode
- Split "De-duplicate image export logic" into "De-duplicate image export" (d535f95), "Add image export fallback" (3576a0d) and "Prevent new tab from grabbing focus" (59a7d02).
- Rewrite `ImageData` as the factory function `composeImageData()`
- Implement `writeToCanvas()`, which creates a canvas loaded with the contents of a dataURL specified by the caller (data encoded via steganography is not lost).
  - Drop support for dataURLs as input to `KC3ImageExport`, since the use-case is no longer valid.
- Modify Imgur upload to use a Blob instead of a dataUrl
  - Drop support for outputting a dataUrl from the canvas, since the use-case is no longer valid.

@DynamicSTOP Turns out the issue with the Blob upload was `$.ajax()` was setting the incorrect `Content-Type`.

ethrundr:
I'll try merging #1960 now.
If Travis persist, then you should debug your own code: there is an error.

Adrymne:
I have no idea what's going on any more, so I'm just going to stop here and hope someone more experienced with git/Travis can figure out what's wrong.

Tedster59:
apparently it worked, so should I merge?

PR #1959, opened 2017-04-06
Sync quests data on different machines via Chrome Storage

reinforce:
This feature allows you to synchronize quests and progress between different computers.
You should enable account synchronization in your browser settings to use this feature.

Note: This feature also synchronizes the quests reset time to avoid a double reset.

Signed-off-by: Aleksei Mamlin <mamlinav@gmail.com>

### DIFF
--- a/src/library/modules/Service.js
+++ b/src/library/modules/Service.js
@@ -483,6 +483,40 @@ See Manifest File [manifest.json] under "background" > "scripts"
 		}
 	});
 	
+	/* On Chrome Storage Changed
+	Sync localStorage parts with Chrome Storage
+	Used for sync quests data on different machines
+	------------------------------------------*/
+	chrome.storage.onChanged.addListener(function(changes, namespace) {
+		// Check if synchronization is enabled, expected changes present and namespace is "sync"
+		if (ConfigManager.chromeSyncQuests && changes.KC3QuestsData && namespace == "sync") {
+			var newValue = changes.KC3QuestsData.newValue;
+			// Check if remote data structure version is matching with current expected structure
+			if (newValue.dataStructVersion == 1) {
+				// Check if local timestamp not set or remote timestamp is more then local
+				if ((typeof localStorage.questsTimeStamp == "undefined") || (newValue.questsTimeStamp > localStorage.questsTimeStamp)) {
+					// Check if JSON strings are the same, then there's nothing to do
+					if (newValue.quests === localStorage.quests) { return true; }
+					// Set new values
+					localStorage.quests = newValue.quests;
+					localStorage.questsTimeStamp = newValue.questsTimeStamp;
+					localStorage.timeToResetDailyQuests = newValue.timeToResetDailyQuests;
+					localStorage.timeToResetWeeklyQuests = newValue.timeToResetWeeklyQuests;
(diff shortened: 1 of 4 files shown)


reinforce:
* Of course computers should be more or less synced in time. Timestamp is GMT, so there is no problem with Time Zones.
* Conflicts should not appear as we have only one game instance at the same time. However, a small chance of conflict can appear because chrome.storage.sync works with delay(about 5-10 sec).

src/library/modules/Service.js
```diff
@@ -483,6 +483,26 @@ See Manifest File [manifest.json] under "background" > "scripts"
 		}
 	});
 	
+	/* On Chrome Storage Changed
+	Sync localStorage parts with Chrome Storage
+	Used for sync quests data on different machines
+	------------------------------------------*/
+	chrome.storage.onChanged.addListener(function(changes, namespace) {
+		for (var key in changes) {
```
Adrymne: Is there a reason you're enumerating over all properties instead of doing something like
`if (namespace === 'sync' && changes.KC3QuestsData) {`
?

reinforce: That was used for checking all keys and changes. Now it can be replaced with such simple check. Thank you.

sinsinpub:
Two suggestions:
 * For backend, add a 'version' number along with timestamp to double check if the data structure in sync storage is matching with current expected structure of quest tracking data. each time if current quest data structure changed and incompatible with old, bump the version number to prevent syncing conflict.
 * For UI, add two things:
   * Add an option setting to disable this feature. because even if ppl login chrome account for syncing, he may not need this feature still.
   * Show a desktop notification (or sth other) when data synced from network (on `storage.onChanged`). no matter success or failure, give ppl a signal to know that local quest data have been overwritten, and safe to start game. notification text may includes the time synced from on success, and on failure, includes the timestamp or version diff to tell the reason why synced fails.

reinforce:
I do not think that the data structure will change frequently, but I'll add checking the versions of the data structure. Is 'vers' would be ok?
As for notification - how about this:
screenshot at 2017-04-09 17 13 25

reinforce:
In which settings category I should place quests synchronization?

reinforce:
@Adrymne, 'vers' used to check version of the data structure, as @sinsinpub wants (if I understood him correctly). If not, I'll rewrite it.

Adrymne:
The issue is that `vers` is only ever initialized to 1 and never changed. Currently, `if (newValue.vers == 1) {` will always be true.

reinforce:
If data structure changed(for example adding yearly quests and reset timer) we manually increase `vers` and rewrite handler to safely overwrite data. To avoid misunderstanding `vers` should be named `dataStructVersion`

Adrymne:
That makes sense. My bad; I was mixing concepts.

reinforce:
Is there any sample for icon? I did not find anything suitable.

sinsinpub:
@reinforce unfortunately no. that's why i introduce a file in `assets/img/quests` folder. in fact all those icons are used for kinds of 'Quests' in game, but there is no icon to represent the whole concept of quests 😅

ethrundr:
Since everything seems fine already, I'll merge this PR

reinforce:
@ethrundr Not so fast, I'll add some changes, which @sinsinpub requests

PR #1963, opened 2017-04-10
Update screenshots management page

sinsinpub:
DONE

* Fix some issues left on #1948
* Update SRoom Screenshots page to manage all types of images uploaded to imgur.com

SS
image

### DIFF
--- a/src/library/modules/ImageExport.js
+++ b/src/library/modules/ImageExport.js
@@ -256,6 +256,25 @@
     return link;
   };
 
+  KC3ImageExport.deleteUpload = function (deletehash) {
+    return new Promise((resolve, reject) => {
+      $.ajax({
+        url: 'https://api.imgur.com/3/image/' + deletehash,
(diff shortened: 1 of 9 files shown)


src/library/modules/ImageExport.js
```diff
@@ -256,6 +256,25 @@
...
+    return new Promise((resolve, reject) => {
+      $.ajax({
+        url: 'https://api.imgur.com/3/image/' + deletehash,
+        method: 'DELETE',
+        headers: {
+          Authorization: 'Client-ID 088cfe6034340b1',
+          Accept: 'application/json',
+        },
+        processData: false,
+        contentType: false,
```
Adrymne: Should be able to remove `processData: false,` and `contentType: false`; they were only necessary to tell jQuery not to mangle the Blob upload.

sinsinpub: done

PR #1965, opened 2017-04-10
Update data synchronization feature

reinforce:
Changes:
* Use 'assets/img/quests/sortie.jpg' as icon
* Use locale-based date and time in notification
* Add a setting for enabling data synchronization notification
* Define version of data structure in QuestManager
* Use data version instead of timestamp for checking new data

Related to #1959 

screenshot at 2017-04-10 19 56 04
screenshot at 2017-04-11 12 08 02

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -775,13 +775,19 @@
 		]
 	},
 	{
-		"section": "SettingsDataSynchronization",
-		"help": "",
-		"contents": [
+		"section" : "SettingsDataSynchronization",
+		"help" : "",
+		"contents" : [
 			{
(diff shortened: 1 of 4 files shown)


reinforce:
The risk is very small, since it takes many years to reach the maximum integer value (2 ^ 53), even if the quests are updated every second.
Let me remind you that the UNIX timestamp(number of seconds since January 1, 1970) will reach the limit in 2038, since it is only (2 ^ 31)

reinforce:
Any other comments?

Tedster59:
looks like everybody likes this, merging.

Issue #1951, opened 2017-04-05
Incorrect Great Success Rate


fourinone41:
Dev panel reports incorrect Great Success rate if mixed sparkled and unsparkled ships are used on a non-drum expedition. Currently it shows (# sparkled ships) x 19%. It should be 0%. 
https://puu.sh/uz0rI/ad7982e14d.jpg
https://puu.sh/uz0vU/519ffe7e0f.jpg

From the [original issue](https://github.com/KC3Kai/KC3Kai/issues/1690#issuecomment-269133444):
> Every ship you sparkle gives you a ~19% chance of GS, ie. 5 sparkled ships gives you a ~95% chance of GS. You are still however, required to sparkle all the ships in the expedition fleet if the expedition does not support overdrumming.

Adrymne:
The error in question is easy enough to patch, but I'd like to make sure there isn't anything else that needs changing before I get started. The current behaviour of the GS chance display, as I understand it, is:

**Regular Expedition**
- If expedition requirements not met, output `???`
- If sparkled ships < 4, output `???`
- If sparkled ships >= 4, output `sparkledShipCount * 19`
  - *Proposed change: If shipCount > sparkled ships, output `0`*
- If sparkled ships >= 4, colour output gold

**Drum Expedition**
- If expedition requirements not met, output `???`
- If sparkled ships < 4, output `???`
- If sparkled ships >= 4
  - If not overdrumming, output `sparkledShipCount * 19`
  - If overdrumming, output `99`
- If overdrumming, colour output gold

I have two concerns. First, `???` seems to be pulling double duty as 'we don't know how to calculate this' and 'can't GS'. Second, the conditions for displaying the gold text seem unintuitive. For example:

Expedition 2, 4 sparkled ships = `~76%` (gold)
Expedition 21, 5 sparkled ships, 3 drums = `~95%` (grey)

(Also, can I get this assigned?)

c933103:
* As I explained on discord, the sparkled GS rate for non drum expedition is not 19n%.
* If the above logic is used then when sparkled ship is 6 it would display more than 100%?
*You can GS with 3 ships too. Not sure about 2 ships.
*Do we have any data on overdrummed but not sparkled GS rate?
*I don't think drum expedition with sparkled ship that are not overdrumed would have that much gs rate

Adrymne:
Forgot to mention GS chance gets capped at 99%.

Incidentally, I assume the discord explanation you're referring to is the following?

[...]

PR #1966, opened 2017-04-11
Improve expedition GS rate prediction

Adrymne:
Changes:

- Fix GS rate for non-drum expeditions with unsparkled ship(s) (see #1951)
- Output GS rate of `"0%"` if expedition requirements are not met (previously was `"???"`)
- Colour GS rate text golden if overdrumming is active (previously would also turn gold when `sparkledCount` >= 4 for regular expeditions)

I am open to advice regarding the latter two changes. The way they behaved previously was bugging me, but I'm uncertain this is better, particularly the last change.

### DIFF
--- a/src/pages/devtools/themes/natsuiro/natsuiro.css
+++ b/src/pages/devtools/themes/natsuiro/natsuiro.css
@@ -1524,8 +1524,17 @@ ABYSS FLEET
 	font-size: 14px;
 	color:#ccc
 }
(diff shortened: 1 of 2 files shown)


Adrymne:
Changelog:

- Implemented KCKai formula for GS rate
- Added more possible colours for GS rate text
- Added + refined tooltips for GS rate

Done working on this unless anyone finds something that needs to be changed.

Tedster59:
has this been updated yet?

PR #1970, opened 2017-04-12
Improved hardcap warning & Added mobile notifications

FOG-Yamato:
Added max Coin constant to PlayerManager

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -366,6 +366,28 @@
 		]
 	},
 	{
+		"section": "SettingsPushAlerts",
+		"help" : "",
+		"contents" :[
+			{
+				"id" : "PushAlerts_enabled",
+				"name" : "SettingsEnablePushAlerts",
+				"type" : "check",
+				"options" : {}
+			},
+			{
+				"id" : "PushAlerts_key",
+				"name" : "SettingsPushAlertsKey",
+				"type" : "long_text",
+				"options" : {"placeholder" : "API Key" },
+				"bound" : {
+					"length_min" : 27,
+					"length_max" : 27
+				}
+			}
+		]
+	},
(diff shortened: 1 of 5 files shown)


FOG-Yamato:
Well without the warning lets ppl be aware of getting close to it so they can change the grind maps and such, but I can't say that it's useful since I never reached the hardcaps myself.

FOG-Yamato:
Added PushBullet notifications, the translation files need a update aswell.

Tedster59:
how does this notification thing work?

FOG-Yamato:
Yes it absolutely does.
image

image

FOG-Yamato:
You have to make an account on PushBullet, create a API key, add it to settings, download the phone app and you're good to go.

Tedster59:
OK, that makes sense.

Nya-chan:
@sinsinpub Pretty sure some of the "you are close" warning are useful - I've personally seen only the coin case (somebody didn't know about hardcap while being close to it and opened lots of boxes, all extra coins lost), but there's no telling for whome else it might be useful.

sinsinpub:
the push notification feature, i break it into another branch: [dev-mobile-push]( https://github.com/KC3Kai/KC3Kai/tree/dev-mobile-push). @FOG-Yamato u could open a new PR. and suggest to write a wiki page to introduce the usage, then add the help link to the settings.

closing warning is okay for me, my point is how those thresholds are defined. eg: warn from max ships - `5`, max slotitems - `20`, they are defined in game. better to have the opinion of person whose resources always near the hardcap, isn't it? so finally, i made it configurable, default value is 95% of hardcap.

FOG-Yamato:
Alright sounds good!

Issue #1962, opened 2017-04-08
Minor equipment showcase text lining problem


Shurichan:
In "full" and "standard" modes, the equipment name and quantity of equipment looks fine.
But, in "light" mode, the text lining looks troublesome.
Happens with both english and japanese equipment names.
![2017_04_08_13_23_55_wcnhnpjfbzahwaaaabjru5erkjggg_1250_1389_](https://cloud.githubusercontent.com/assets/13228415/24828563/747e89b6-1c68-11e7-9ae0-70359ae5fed4.png)

DynamicSTOP:
i would say it's not a mode problem but a single line equip text.
![default](https://cloud.githubusercontent.com/assets/9434504/24829563/7234df78-1c74-11e7-8acd-9bbe23b124df.png)

DynamicSTOP:
how about this?

![default](https://cloud.githubusercontent.com/assets/9434504/24829849/f9b9dffc-1c79-11e7-8583-ec12d1562aad.png)

DynamicSTOP:
Ended up with this two variants [Jp](http://i.imgur.com/cmmkfgg.png) [Eng](http://i.imgur.com/Al3MNkB.png)

sinsinpub:
@DynamicSTOP no such necessary... because Japanese fonts installed on your (other's) system might be different, it's not possible to fit with only a little width 😂 
eg, my output (font auto mapped by system ofc) will need more space:
![image](https://cloud.githubusercontent.com/assets/160240/24830955/8024c6cc-1cc2-11e7-888c-07e0cee55518.png)

DynamicSTOP:
Noooooooo...... then i need to move everything inside (...) on new line.

DynamicSTOP:
Fix merged into master. Closing this.

PR #1972, opened 2017-04-12
single line alignment and split for long words

DynamicSTOP:
this is how split looks
default
and this is single line example
default

compared to how it was looking before
default


DynamicSTOP:
#1962

Tedster59:
nice, merging

PR #1974, opened 2017-04-13
Cumulative Update

sinsinpub:


Adrymne:
https://github.com/KC3Kai/kc3-translations/commit/45a7374cf21e4c4f3f91d63c1c7e869d4519e8fe will cause the GS tooltip to break until #1966 is merged. Specifically, drum expeditions will have `ExpedGSRateExplainExtraDrum` show up as the third line.

(In hindsight I probably should have just reused that key...)

Issue #1989, opened 2017-04-23
Unwanted scroll bar


ehtnnd:
the Top Margin setting if number is more than 0 a scroll bar will appear

http://i.imgur.com/ouGIGaC.png

http://i.imgur.com/19SMmOH.png

### Steps to reproduce: see above

Issue #1824, opened 2017-02-13
Positioing of background picture is broken


Nya-chan:
Please make sure that you have these details on hand when submitting your issue/bug report as this will greatly speed up the issue triage process -- thank you!

- [x] Your Chrome version - 56.0.2924.87
- [x] Your OS and its version - Win 10
- [x] 32bit or 64bit? 64bit
- [x] Your KC3 Kai version - 31.0.6
- [x] Which build you have tried running? Webstore version stable
- [x] Screenshots which accurately paint the issue(s) you are facing. - See below
- [x] A step by step method that you did to describe how you got to that issue.

### Steps to reproduce:
- Enable a background with a background-position set to 100% of Y position or similar (calc based off of 100 %, etc).

### Actual behaviour:
- The background gets shifted upwards, ending way above the frame

### Expected behaviour:
- The background should stay where it's told to by background-position property, with 100 % being the bottom of the window

### Screenshots (if applicable):
https://i.imgur.com/HlCkS4d.png
https://i.imgur.com/k3vNg31.png

sinsinpub:
@dragonjet is it still counted as bug...? 😅  i've already tell this issue at PR: https://github.com/KC3Kai/KC3Kai/pull/1722#issuecomment-274439407

dragonjet:
@sinsinpub I was thinking we can workaround something on DMM to make it work correctly. I have failed to do it once in local, but haven't given up that we can make the DMM page background to work the same way as any other page

PR #1996, opened 2017-04-25
Hotfix 31.5.2

Vultren:
### ⚓️  Enhancements & Fixes

* #1992 Improve Akashi Repair timer display
  * Move repair time tooltip down
  * Fix Akashi repair timer output if start time is unknown
* #1991 Update styles for DMM site play mode
  * Fix background image size not cover issue (#1824)
  * Take top margin and zoom scale into account (#1989)
  * Add a setting to customize CSS of DMM top page

### ⚓️  Translations & Data

* Add 4th Anniversary Lines
* `en` quotes
* `en` terms
* `jp` terms
* `th` quotes
* `ru` ships


sinsinpub:
why not follow `fixes -> master -> webstore` flow...? 😓 
btw, also push the workaround to user on #1991

Vultren:
Well I didn't want to mess up a hotfix by adding extra stuff again and this was only for the limited seasonal lines. normally I would do it the normal way of master -> webstore

sinsinpub:
only update TL pointer also possibly mess up by extra stuff :joy:

Vultren:
All this really does is update the pointer. well close this if you want and we'll push it along with the CSS and akashi timer

PR #1999, opened 2017-04-26
Show green border around new ship drops

Tibo442:
Kind of a reminder that the dropped ship is a ship you don't have and you *might* want to heart lock her.

Possible 'bug': If someone gets a new ship on a early node, and then get the same ship in a further node, it will get tagged as a 'new' ship. This is because ships are added to `KC3ShipManager` **after** returning to port.

Visual example

### DIFF
--- a/src/pages/devtools/themes/natsuiro/natsuiro.js
+++ b/src/pages/devtools/themes/natsuiro/natsuiro.js
@@ -796,7 +796,7 @@
 		$(".module.activity .battle_rating img").attr("src", "../../../../assets/img/ui/dark_rating.png").css("opacity", "");
 		$(".module.activity .battle_rating").lazyInitTooltip();
 		$(".module.activity .battle_drop img").attr("src", "../../../../assets/img/ui/dark_shipdrop.png");
-		$(".module.activity .battle_drop").removeData("masterId").off("dblclick");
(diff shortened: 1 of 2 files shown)


src/pages/devtools/themes/natsuiro/natsuiro.js
```diff
@@ -2253,6 +2253,7 @@
 						.data("masterId", thisNode.drop)
 						.on("dblclick", this.shipDoubleClickFunction)
 						.attr("title", KC3Meta.shipName( KC3Master.ship(thisNode.drop).api_name ))
+						.toggleClass("new_ship", KC3ShipManager.find(w => RemodelDb.remodelGroup(thisNode.drop) === RemodelDb.remodelGroup(w.masterId)).length === 0)
```
sinsinpub: a little complain: this expression is brainfucker 😂

Tibo442: `KC3ShipManager.find(w => RemodelDb.remodelGroup(thisNode.drop) === RemodelDb.remodelGroup(w.masterId))`

Selects all ships that are in the same remodel group as the dropped ship, eg dupes. Feel free to add a comment above/below to explain it for future maintenance.

sinsinpub: nvm, already comment it at f60a29b 😃

PR #2009, opened 2017-04-30
Add experience/level history

Tibo442:
Using radio buttons instead of checkboxes unlike resources/consumables because scale of xp compared to level is huge.


sinsinpub:
i think, for more useful, should also record rank points, especially for extra pts from monthly EO (and Op Z either).

Tedster59:
agreed there

PR #2014, opened 2017-05-02
Battle replay bossbar

Vultren:
Allows the battle replay to show the boss bar:
 * For regular world maps (including EO maps): only new runs will record boss bar info, old replays will not show.
 * For event world maps: new runs ofc, if old maps stats data not lost, old replays will also show boss hp bar.

### DIFF
--- a/src/library/managers/SortieManager.js
+++ b/src/library/managers/SortieManager.js
@@ -67,6 +67,14 @@ Stores and manages states and functions during sortie of fleets (including PvP b
 				time: stime
 			};
 			// Add optional properties
+			// Record states of unclear normal (or EO) maps
+			if((world < 10 && mapnum > 4) || typeof thisMap.kills !== "undefined"){
+				sortie.mapinfo = { "api_cleared": thisMap.clear };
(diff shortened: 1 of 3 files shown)


sinsinpub:
Two things;
* map gauge info should locate after `sortieData.battles.length === 0` condition
* handle all kinds of possible `undefined` exception

sinsinpub:
Basically works now. For event HP bar, will try to use the boss end hp of previous sortie which means start hp of this sortie. For normal kills bar, replayer should change the handling on `defeat_count` >= max kills, should be empty hp bar (or just hide it) @fourinone41

mikenzb:
I would like to merge however monthly EOs aren't displaying the correct HP atm, the replayer takes the final battle bar HP as it's starting HP, as soon as that's fixed I'll merge.

sinsinpub:
To finally solve history EO issue... normalize the hp gauge info source now:
Like the event map HP data, will also record the `cleared` and `default_count` states for normal (EO) maps into sortie database table, and only use them instead of inaccurate `localStorage.maps` data. As a result, all old history replays will not contain `default_count`.

PR #2032, opened 2017-05-06
Resolve quest sync conflicts with merge

Adrymne:
* When loading quest data from chrome sync, merge it with local data instead of discarding.
  * Discard expired repeatable quests when merging
* Trigger sync immediately when user enables chrome synchronization.
* Remove repeatable quest reset times (e.g. timeToResetDailyQuests) from `KC3QuestManager` instance; they are now stored solely in localStorage.

Note: the merge logic is currently insufficiently sophisticated to give the correct answer in some cases. Consider the following scenario:

1. 2/10 on machine A
2. Sync progress on machine A
3. 1/10 on machine B
4. Enable sync on machine B
5. 2/10 on machine B

Simple adding the two together will not (always) produce the correct result either; some of the progress on machine A might have originated on machine B. The auto-adjust muddies the issue further.


Adrymne:
Changelog:

* Added `KC3Network.trigger()` call

Adrymne:
Found a bug - sometimes local changes will cause the sync notification to be fired. Currently investigating the cause.

Adrymne:
Changelog:

* Move the sync trigger logic into its own module.
* Stop listening for remote changes when local data is written (remote data is lost at this point anyway)
* (Re)start listening when the `QuestSync` module is reloaded.

Adrymne:
Changelog:

* Resolve notification icon's path with `chrome.extension.getURL()`

Adrymne:
Changelog:

* Update Gruntfile for new QuestSync files
* Prevent the `console.error()` calls in `KC3QuestManager.logError()` and `KC3QuestSync.logError()` from being removed
* Fix `KC3QuestSync.init()` not being called (should now be called when devtools panel is initialized).
* Load latest config before checking whether quest sync is enabled inside `KC3QuestManager.save()`

Adrymne:
* Fixed merge conflict (updated for new monthly)
* Added KC3QuestSync to `.jshintrc`

PR #2079, opened 2017-05-27
Add logging module and SR tab

Adrymne:
* New module `KC3Log` to aggregate logging in a context-agnostic manner.
  * Hooks into `console.*` methods
  * Can call the `error()`/`warn()`/`log()`/`info()` methods directly if you need to do something after the logging completes
* New dev-only SR tab "Logging" to display logs from all contexts.
  * Click on a log entry to view additional data (including stack trace from errors).
* New export on SR tab "Profile" to dump logs to JSON.


Adrymne:
@sinsinpub Both, as you surmised. Having all logging in one place is convenient for devs and I suspect end users will be less likely to balk at being asked to open up an SR page as opposed to several consoles.

(General note: I won't be rebasing immediately to make changes easier to see; please let me know before merging so I can tidy up.)

Adrymne:
> too many noise logs saveLogToDb caused by chrome.runtime.onMessage listener...

This one's easy enough to fix, I can just define a new onMessage listener rather than using the one in Service.js.

> will be nice if also save and show the logging from source location file.js:lineNo

This is a bit trickier, but it should be possible to extract the information from a stack trace if nothing else. I suspect there won't be a way to make the link in the chrome console point at the call site, though.

sinsinpub:
FYI, because the `WhereClause.anyOf()` of dexie.js will cause sort order of result collection undefined (see [`.or()` doc](http://dexie.org/docs/Collection/Collection.or().html)), only sortBy limited records not enough, different log types among pages will be in unexpected order. I tried this, should solve:

```javascript
	return this.con.logs
		.orderBy('timestamp')
		.reverse()
		.and(entry => typesToShow.indexOf(entry.type) !== -1)
		.offset((pageNumber - 1) * itemsPerPage).limit(itemsPerPage)
		.toArray();
```

Adrymne:
I had some local changes along the same lines, so I ended up combining things manually. I still need to reimplement `console.assert`; `lookupErrorObject` was a casualty of the rewrite as well.

Also, the native console methods are now exposed at `KC3Log.console.*`.

sinsinpub:
that's okay. i just want to ask about assert and non 1st Error argument situation lol

another known issue just wanna remark (but may not need to be solved) about saving stack for all levels and do not save as `source` field: because stack of `console.error` caller will be replaced by `Error.stack` (if there is one) which is the stack of exception thrown point, the callsite source will not indicate the line of `console.error` as non-Error cases, but the exception thrown line instead. it's indeed not important because we usually concern on the Error, not the `console.error` line itself.

Adrymne:
That's everything (I hope). Unless anyone has something that needs changing, I'll rebase tomorrow.

PR #2104, opened 2017-06-14
Refactoring ship list based pages

sinsinpub:
Just wanna reuse more codes for those `ship list` liked pages, current: `docking`, `shipquests`.
Should no feature changed. More similar pages will be easier to be created.
(rebased on `update-cumulative` branch)


sinsinpub:
not into master directly, close this

PR #2154, opened 2017-08-07
Update certain assets on the fly

Tibowl:
Overrides jQuery attribute function to intercept ship images. Saves files to IndexedDB for caching.

Bugs for on-the-fly added ship: missing shipboxes flash when updated.
Bugs for on-the-fly meta: if some code accesses meta before changes loaded from cache/from github -> uses old data. Tested in strat room (exp calc), reloading causes it to use old data, clicking on refresh tab or switching tabs uses changed data. 
Not really a problem in dev tools since it waits for server packets.

Only checks for updates to github's assets when popup gets opened.


src/library/managers/ImageManager.js
```diff
@@ -0,0 +1,96 @@
...
+	window.KC3ImageManager = {
+		errorImage : "../../../../assets/img/ui/empty.png",
+		cdn : "",
+		//imageCache : {},
+
+		init :function() {
+			var self = this;
+			var originalAttr = jQuery.fn.attr;
+
+			self.cdn = "https://raw.githubusercontent.com/KC3Kai/KC3Kai/" + (localStorage.lastestCommit || "1809dc3f110f694c52f9d89dc9404c726ef1f02e") + "/src/";
```
sinsinpub: default commit can be `master` or `webstore`

src/library/managers/ImageManager.js
```diff
@@ -0,0 +1,96 @@
...
+			self.checkForUpdate();
+
+			jQuery.fn.attr = function() {
+				var arg = arguments;
+				var jQueryObject = this;
+
+				if(arguments.length > 1 && arguments[0] === "src") {
+					var url = arg[1].replace(/(\.\.\/)+/g, "");
```
sinsinpub: might double check: `src` starts with `chrome-extension://` and (or) `/assets/img`

Gruntfile.js
```diff
@@ -320,6 +320,7 @@ module.exports = function(grunt) {
 				src: [
 					'src/manifest.json',
 					'src/data/*.json',
+					'!src/data/WhoCallsTheFleet*.json',
```
sinsinpub: oh, i forgot to mention: should rename extension of `WhoCallsTheFleet*` from `.json` to `.nedb` (as original author also done). because they are not standard json, they are: https://github.com/louischatriot/nedb
by renaming, will not need to exclude them from jsonlint, just add them to copy:processed session.

Tibo442:
Made a couple changes while rebasing with latest update: fixed duplicate caches for filenames with start / and without. And fixed the ship avatar override not working on equipment list page.
Am kinda forced into running this fork for almost half a year since it changes my db version, making it rather difficult with making other PRs (have to test on branch with this stuff, make pr without). Haven't had any issues with stuff of this pr in that time so it's pretty stable. Wonder if it can make into webstore before event so we can have live updating for edges/new ship cgs.

There are still a couple minor problems that might need some discussion regarding live updating of seasonal cgs: currently it'll only get cgs of ships that went from no cg to some cg. It won't get changed cgs from remote since it has them locally. 
Also only update method of assetsversion is still just opening of extension menu. Might run something in panel on a fairly long timer ~1 hour?

Issue #2060, opened 2017-05-14
Replay Image's battle data too long for steganography


fourinone41:
The battle data embedded into the replay image appears to have a max length of 30000 (?). Anything longer is truncated. This becomes a problem for longer sorties.

Example: https://cdn.discordapp.com/attachments/307882092287623188/313209192607514624/23725.png
Returned by `steganography.decode`: https://pastebin.com/raw/deq6udqx

We may need to consider reducing the length of data stored.
Can potentially remove "airRaid" data; replayer does not currently use it and I'm most likely not going to implement it.

sinsinpub:
🤔 seems enlarging image size also increases the size of alpha channel,  which might contain more encoded data...

fourinone41:
I saw you cleared `"lbas"`, I actually want to use that for Show Fleet Details (just implemented it yesterday before seeing this issue actually hah), okay if you bring it back? Unless it's still too long.

Also `"airRaid"` is not on `sortieData`, it's on each `sortieData.battles`.

Vultren:
How to the air raids on non-battle nodes get saved anyway?

sinsinpub:
@fourinone41 fixed. also cleared `battle.shizunde`, i think it's also not used?

sinsinpub:
@Vultren no place to save it anyway. because it's in the api `/next`, which we never consider to save. there was a discussion about creating virtual nodes. but i'm lazy, just saving it along with battle node😂

Vultren:
`battle.shizunde` is defiantly never used. I have never seen that change at all, unless it is for allies only

PR #2170, opened 2017-08-17
Scale steganography image with large sortie data

Tibo442:
Changing steganography.js's "t" (bits per pixel) breaks decoding of old images which might cause undesired effects on battle replayer.


sinsinpub:
related to #2060

PR #2185, opened 2017-08-23
Showcase update: event ship locks

DynamicSTOP:
image
i guess i should ask someone to validate my engrish :rofl:

so it looks like so
first you either lock your ships in sortie or inside strategy room **Locking** page.
image
 
Then you click that checkbox above and have something like this
image

or in case of Light version this
image

Colors might be discussed but i rather leave them as is, since Tanaka already trolled us with this two:
image


DynamicSTOP:
Luck will look like so:
default
0-39 -> nothing
40-49 -> half bar
50+ -> full bar

PR #2198, opened 2017-08-28
Fix quest sync fail when data is too large

Adrymne:
Pulled the new wrapper out into its own module. Implementation is largely based on [this](https://github.com/kdzwinel/Context/blob/master/js/classes/HugeStorageSync.class.js), with a couple of differences:

* Old chunks that are not overwritten are instead set to `null`. This potentially saves some space if the stored value's size shrinks.
* Fix chunk size calculation: `chrome.storage.sync.set` will stringify the value it receives, which can cause its size to increase.


Adrymne:
As per the [docs](https://developer.chrome.com/apps/storage#property-sync):

> The maximum size (in bytes) of each individual item in sync storage, as measured by **the JSON stringification of its value plus its key length.** 

I think [this](https://cs.chromium.org/chromium/src/extensions/browser/api/storage/settings_storage_quota_enforcer.cc?q=set+file:%5Esrc/extensions/browser/api/storage/+package:%5Echromium$&dr=CSs&l=39) is the relevant part of the chromium source.

(Updated the comment on `generateChunk()` to reflect this, as well as removing an old comment about usage.)

PR #2213, opened 2017-09-08
strategy room locking page update

DynamicSTOP:
Replaced tiny divs with shiplistgrid table
Added some basic filters


DynamicSTOP:
Drag&Drop ship icon from table or click inside tag column. 
Only ships w\o sally( not locked in game) will get planned lock
Double click on icon in box to clear or clear all by clicking on div at top right

PR #2219, opened 2017-09-11
crossed screws image in case player don't have Akashi yet

DynamicSTOP:
image


DynamicSTOP:
just in case, 500 here is because i made backup from outdated version which doesn't have caps in `localstorage.player`

PR #2221, opened 2017-09-12
Add mark that show available quests in Required Ships tab

reinforce:
* Fix few ship id in ship_quests.json
* Add mark that show available quests in Required Ships tab
default


sinsinpub:
\oh and there are new quests/ 😃

PR #2238, opened 2017-09-23
Fix large searchlight F33 LoS

Tibo442:
Pre fix: deckbuilder vs kc3

Post fix: deckbuilder vs kc3


sinsinpub:
i just updated this method too, get rid of all those 0.6... future unknown los equipment will get 0.6 by default.

PR #2249, opened 2017-09-29
seasonal icons

nonsensed:
Work in progress -- I still need to find the new CGs


sinsinpub:
aren't new cgs in sroom update page...?

nonsensed:
Among other things, I may be slightly paranoid about burst-loading assets from the server. Also, my current internet connection is too lossy to load most of those. Not really sure about cropping files from there.

PR #2251, opened 2017-09-30
Change FCM to fish counter during fish events + split mamiya & irako in tooltip.

Tibo442:
When player has at least 1 mackerel: show mackerel counter
When player has 0 mackerel (eg, outside of events): show FCM counter

Also split up mamiya and irako counter in hover

image

### DIFF
--- a/src/pages/devtools/themes/natsuiro/natsuiro.html
+++ b/src/pages/devtools/themes/natsuiro/natsuiro.html
@@ -236,7 +236,7 @@
 							</div>
 							<div class="consumable page3">
 								<div class="consumable_icon"><img src="../../../../assets/img/useitems/61.png" /></div>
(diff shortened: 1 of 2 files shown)


sinsinpub:
a suggestion: might replace the 1st class medal instead? it seems also an event thing, and 'useless' :joy:

nonsensed:
This is just a side note and it doesn't really matter, but we do have some inconsistency between referring to the fish as `Saury` or `Mackerel` or `Mackerel Pike`. Might have to take a close look later on.

I use `saury` in translations because it's more accurate as far as I know.

sinsinpub:
@nonsensed u could just update them in `useitems.json`. don't care internal naming 😂

nonsensed:
Yeah, I know. xD I mentioned it for a slightly different reason.

PR #2256, opened 2017-10-01
fix strategy room showing max modernization value of ship stats when …

MegaAbsol:
…it should be showing the current stat

image
Here's an example of what it was like previously. For example, Zuikaku only has a FP value of 0, yet the game displays 39, the maximum stat.

ValuesTuple is stored like so:
`[maxmod value, current stat value, current stat + equips]`

Code used to be:
```javascript
$(".stat_value", statElm).text(valuesTuple[equipStatIndex > 0 ? (this.equipMode ? equipStatIndex : 0) : 0]);
```
which would return either index 0 or 1 (zero if no equips, 1 if show equips is selected). This resulted in displaying of maxmod value if equip stats was off:
equip stats off
and displaying of base value without equips if equip stats were on.
equip stats on

The changes in the code aims to fix this by adding 1 to the index. I'm not sure if there's another underlying issue that was the root of this whole problem, but this seems to work:

image

image

### DIFF
--- a/src/pages/strategy/tabs/ships/ships.js
+++ b/src/pages/strategy/tabs/ships/ships.js
@@ -1100,7 +1100,7 @@
 		--------------------------------------------*/
 		modernizableStat :function(statAbbr, rowElm, valuesTuple, equipStatIndex = 1, isSup = false){
 			const statElm = $(".ship_" + statAbbr, rowElm);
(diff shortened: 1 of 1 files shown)


MegaAbsol:
glad to be of help 👍

PR #2283, opened 2017-10-22
Readd pages + fix alternating in ships list when filtering by name

Tibo442:


Tibo442:
For 250 ships, searching 'a' (210 matches); before this patch ~200ms, with this; ~320ms

PR #2299, opened 2017-11-02
Add ability to time subtitles

Tibo442:
Allows per line timing of subs. Also allows custom fadeout times for slow lines.
Example for Yamagumo:
```json
        "Poke(2)": {"0,2300": "You can't do that~ That's delicate~ The depth charges, they're sensitive~"},
```

Example for quotes.json with one of the new quests; https://www.youtube.com/watch?v=VU8S80Bghqs
```json
        "1187": {
            "0":"Michishio: \"Why did they assign me to this fleet...?\"",
            "3530": "Shigure: \"Oh, Michishio. Nice to see you.\"",
            "7500": "Michishio: \"Shigure?! What are you doing here? Don't tell me we're in the same unit!\"",
            "15720": "Shigure: \"Looks like it... Force C of the First Striking Force...\"",
            "22900": "Michishio: \"What's wrong, Shigure? You don't look so good. Maybe you should lie down for a bit.\"",
            "30540": "Shigure: \"It's okay, I'm fine. It's because of the rain... I guess. Anyway, let's go.\""
        },
```

### DIFF
--- a/src/library/injections/dmm_takeover.js
+++ b/src/library/injections/dmm_takeover.js
@@ -286,7 +286,7 @@
 		/* SUBTITLES OVERLAY
 		Displays subtitles on voice audio file requested
 		--------------------------------------*/
-		subtitleVanishTimer: null,
+		subtitleTimer: null,
 		subtitleVanishBaseMillis: null,
 		subtitleVanishExtraMillisPerChar: null,
 		subtitleHourlyTimer: null,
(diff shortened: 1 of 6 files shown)


sinsinpub:
to define a timeline of subtitles, 'delay' might be `delayShown[, delayVanished]` format (vanish one optional, vanish at next line shown or a max value by default)
so that no trailing empty line "" needed...

sinsinpub:
to clarify, points about delay times of current implementation are:
 * delay values of multi-lines are in milliseconds, relative to the first line appearing. for each value, it controls both timing of current line showing and previous line hiding. not possible to define hiding timing of previous line individually.
 * the vanish milliseconds in `,vanishms` only affects the last line. if it omitted, will use default algorithm to get a time as usual single line quote does, which using the text length of last line. the values defined at non-last lines will be ignored.
 * iron out that if keys of an object is in pure number (integer), they will be auto sorted in ascend order. so make sure delay times are ascend with the same order of lines. this also means using `delayms,vanishms` in non-last lines might break the order.

maybe i didn't make my idea clear, but current implementation is okay for me 😅

PR #2302, opened 2017-11-03
Add previously encountered abyssal fleets to battle activity

Tibo442:
Show previously encountered fleets to compass.

image
Might need some more css tweaks.

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -142,6 +142,12 @@
 				"options" : {}
 			},
 			{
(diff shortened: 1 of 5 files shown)


Tibo442:
`(encounters elements will appear and mess up battle activity).`
Shouldn't happen since the `node_type_prev_encounters` gets hidden when the `node_type_battle` gets shown. It *could* cause some stutter when it does load in the background.

PR #2306, opened 2017-11-05
Routing related DB Submission

Tsubakura:
TsunShip joins the fray PraiseTheTsun

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -28,6 +28,7 @@
 		"KC3Meta": true,
 		"KC3Network": true,
 		"OpenDBSubmission": true,
+		"TsunDBSubmission": true,
 		"KC3Panel": true,
 		"PictureBook": true,
 		"PoiDBSubmission": true,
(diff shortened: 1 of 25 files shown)


Tsubakura:
I dont mind at all, edit away until it works. :>

sinsinpub:
remark some changes i made for submission module:
 * Use api param instead of  `KC3SortieManager.fleetSent`. should not try to use data from other managers/modules as possible as we can, because data submission is executed before API processed by other part of KC3 modules, data might be out of date.
 * Fix fleet data collector: eLoS F33 function name, condition of `-1` ship, equipment from ex-slot. Question: stype() is a translated name of ship type, not its ID, do u really need ship types in different languages?
 * Add some properties to data, such as `debuffSound` we metioned; `api_event_id` & `api_event_kind` to indicate real node type (`api_color_no` is not node type)
   * A suggestion: if we are not sure which properties of API data to be collected for some part, why not just send raw api result with all properties? because after event started, updating the data parsing codes will cause a time-lag and useful data-loss.
 * Take care of `undefined` ref error as always do

PR #2308, opened 2017-11-06
Routing database submission

Tsubakura:
* Including PRs:
  * https://github.com/Tsubakura/KC3Kai/pull/1
  * https://github.com/Tsubakura/KC3Kai/pull/2
* Replace n/a KC3 data submission setting with new one

To answer sin_sin's questions:

- Preferably, I dont want to use stype() at all, I just need to know the ship type id.
- Wouldnt it take more time to send everything? Is it also necessary? The data already includes anything that is sortie related, excluding LBAS setups.

I prefer that those properties in data are added back in during the initialization, it serves as a summary as to what data is sent over to the server.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -28,6 +28,7 @@
 		"KC3Meta": true,
 		"KC3Network": true,
 		"OpenDBSubmission": true,
+		"TsunDBSubmission": true,
 		"KC3Panel": true,
 		"PictureBook": true,
 		"PoiDBSubmission": true,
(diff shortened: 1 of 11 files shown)


sinsinpub:
retrieve latest data from ship instances might be ok, since `/api_get_member/ship_deck` is called before `/api_req_map/next`.

i guess u should still submit ship master id (or `api_id`), not only the `api_name`. and for equipment, might also add their `api_name`.
well, all these data requirements are depending on ur implementation of report server 😸

PR #2344, opened 2017-11-20
Open fleet manager from sortie/event page

HitomaruKonpaku:
Click on the sortie id now will load the fleet manager
Right click will open in the new tab


HitomaruKonpaku:
Hmm sorry for the file format I'll recheck it

And should i just remove the right click event?

HitomaruKonpaku:
Is it ok now?

Adrymne:
It does now work from both Maps and Events, which is great. However, ctrl+click is treated as a right-click on OSX, so open-in-new-tab will break. In general, OSX uses cmd instead of ctrl.

HitomaruKonpaku:
@Adrymne so instead of ctrl, alt will be better?

Adrymne:
ctrl is fine - open in new tab is expected behaviour. You just want to listen for the cmd key as well (reference: [Mousetrap](https://github.com/ccampbell/mousetrap/blob/master/mousetrap.js#L135)).

Issue #2335, opened 2017-11-19
PoiDBSubmission problem


gakada:
I don't know if there is already an issue/fix, but [PoiDBSubmission](https://github.com/KC3Kai/KC3Kai/blob/master/src/library/modules/PoiDBSubmission.js) does `api_ship_ke.slice(1,7)`, which is now incorrect. Poi reporter was [updated](https://github.com/poooi/plugin-report/commit/843702876444435134d5f8d93c2c0f59ff0b5bd6) to do
```
drop.enemyShips1 = body.api_ship_ke
drop.enemyShips2 = body.api_ship_ke_combined
```
so not `enemyShips`, but `enemyShips1` and `enemyShips2 `, always sending `api_ship_ke` and `api_ship_ke_combined` as is.

PR #2374, opened 2017-11-29
Update poiDBSubmission.js

sorewachigauyo:
As mentioned in #2335 , updated poiDBSubmission.js to reflect new changes in poi drop ship record model

Added night-to-day battles 
Changed enemyShips to enemyShips1 (for enemy main fleet) and enemyShips2 (for enemy combined fleet escort, if any) as of https://github.com/poooi/plugin-report/commit/843702876444435134d5f8d93c2c0f59ff0b5bd6
Removed slice for enemy comp data, send data as is

EDIT: Removed some api data as poi reporter does not seem to use them

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "KC3Kai",
-  "version": "31.11.4",
+  "version": "31.11.5",
   "description": "Kantai Collection Game Viewer and Tools",
   "repository": {
     "type": "git",
--- a/src/library/managers/GearManager.js
+++ b/src/library/managers/GearManager.js
@@ -15,7 +15,7 @@ Saves and loads list to and from localStorage
 		// These IDs can be updated at `fud_weekly.json`
 		carrierBasedAircraftType3Ids: [6,7,8,9,10,21,22,33,39,40,43,45,46],
 		// Dupe `api_cost`, `api_distance` fixed for non aircraft gears since 2017-03-17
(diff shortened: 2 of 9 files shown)


gakada:
`enemyShips1.length` and `enemyShips2.length` don't need to be 6. Can it be
```js
dropShipData.enemyShips1 = response.api_ship_ke
dropShipData.enemyShips2 = response.api_ship_ke_combined
dropShipData.enemyFormation = response.api_formation[1]
```
instead of https://github.com/KC3Kai/KC3Kai/blob/7cfb108a4b910d0f82469d12eb77b1ab458b0d4a/src/library/modules/PoiDBSubmission.js#L172-L204 setting formation to 0 or sending -1 will make KC3Kai special, if it fails, let the handler fail, enemyFormation will be undefined, enemyShips1 and enemyShips2 will be [], like in poi reporter. Also maybe https://github.com/KC3Kai/KC3Kai/blob/7cfb108a4b910d0f82469d12eb77b1ab458b0d4a/src/library/modules/PoiDBSubmission.js#L231-L234 can be removed? (Is `api_ship_id` correctly represent `api_ship_ke` and `api_ship_ke_combined`? Do we need it, when report is already broken, since there is no formation? More importantly, why target a case that shouldn't happen normally.)

mikenzb:
PR remade in #2381

PR #2391, opened 2017-12-04
Update support expedition estimation based on new criteria

sorewachigauyo:
Based rules from https://twitter.com/Lambda39/status/934719589740527617


sorewachigauyo:
Still being researched, likely conditions are 1CVL+1 CAV/BBV/CV/AV/AO/AS/LHA/SSV/CL (there should also be BBV/CV limit)+2DD

Probably need slot checking
Need equipment tracking with ASW stat for CVL and aviation ship, like autogyro/t3 liason/SPF/SPB/carrier bomber/seaplane recon

PR #2398, opened 2017-12-11
Added exporting current fleets for winter 2018 event

DynamicSTOP:


DynamicSTOP:
image
should make something like this

DynamicSTOP:
Probably will be die once Winter event would be over. May be earlier.

ethrundr:
On the other hand, if we can make it dynamic enough (like, modify the fleet name and ships in it), we can reuse it on the following event, to let players prepare and share their fleet with upcoming events.

Just think this as enhancement to current shiplock menu, but instead of looking for tags, we configure it manually as for who in which fleet.

DynamicSTOP:
>maybe host base image in our repo? (ref github cdn)  

`GitHub Pages sites have a soft bandwidth limit of 100GB per month.`
I thought it's 1GB for some reason :thinking: Not sure where to place in github.

sinsinpub:
@DynamicSTOP here `/src/assets/img`

Adrymne:
Don't know if JPG compression is a concern here, but if it is, you can pass `quality` or `format` props to `KC3ImageExport`'s constructor.

DynamicSTOP:
Guess compression isn't a big concern. I don't think that it's more important then any showcase export and for that one we take same as for screenshots :thinking: in the end the most important things are numbers and remodel text here instead of perfect pixel image.

omoito:
Why do the shipgirl icons look like they were ran through waifu2x? All the faces are awfully blurred.

DynamicSTOP:
@omoito 
Image should use same extension as screenshots that you've selected in KC3settings. Try `png` there.
Although if you consider [this one](https://raw.githubusercontent.com/KC3Kai/KC3Kai/master/src/assets/img/ui/event_winter_2018.png) as blurry then there is not much you can change.

omoito:
@DynamicSTOP 
That's definitely blurry. The best example is Hatsuharu, whose hair has become a purple smudge. And then there's also Yamashiro whose mouth is gone. I assume the point was to get rid of artifacts but in the process it ended up killing quite a bit of detail. The Japanese text also gained some blurriness.

PR #2405, opened 2017-12-13
Format panel ship left exp, resources

Tibo442:
Format ship left exp, resources and furniture coins.

Result depends on user's locale;
image


sinsinpub:
i guess not everyone like thousands separator, even in-game numbers (like exp) has no separator...

PR #2441, opened 2018-01-25
Override DMM reload dialog, making it only show once per session.

Tibo442:


Tibo442:
Can't access `window.DMM` from content scripts, need to inject into page to access it. And without the manifest, won't be able to load it from a file.

Could either put it into cookie.js for all games (and adjust injectable for sites without window.dmm) or something else to inject it on `dmm.com/netgame/*`.

PR #2485, opened 2018-03-02
Allow to filter ship list by RegEx

HitomaruKonpaku:


sinsinpub:
btw, personally i don't care about 'regex' or 'includes', just to remind: test and confirm that its performance  is acceptable for guys like u owned 300-slots. if not, another check box to toggle match method between 'regex' and 'includes' might be the solution.

HitomaruKonpaku:
Ya already tested, I don't have any performance issue

PR #2486, opened 2018-03-03
Fix battle prediction bug for partial CF

Adrymne:
`Kouku.extractDamageArrays` did not pad the damage arrays before merging;
this caused issues as KCSAPI truncates the arrays if ships are missing,
and the array elements' indices are used to determine attack target.

Concretely, arrays `[1, 2, 3]` and `[6, 7]` were being output as `[1, 2, 3,
6, 7]` instead of `[1, 2, 3, 0, 0, 0, 6, 7, 0, 0, 0, 0]`.

Example of bug: [#1](https://cdn.discordapp.com/attachments/205766705463427072/419041544562278401/KC3-Error-1519977512.json)


sinsinpub:
oh, ive done the same fix at `update-cumulative`, could u rebase on that and refine it?

PR #2499, opened 2018-03-16
Misc. updates

FOG-Yamato:
Updated Readme
Updated About page
Removed KC3 and Kancolle-DB submission modules and mentions


FOG-Yamato:
I mean I guess we can leave them since we're gonna have to scrap everything anyway in 2 months

PR #2522, opened 2018-03-27
* rumored spring event ships

DynamicSTOP:


DynamicSTOP:
it's just a habit from work, start everything from * so it will be like unsigned list.

PR #2529, opened 2018-04-04
 Add trigger chance for 12cm 30tube Rocket Launcher Kai Ni (Rosa K2)

sorewachigauyo:
From Discord, tentative formula
```
Proc chance = (base AA stat + luck + adjusted equipment AA) / ( 400 - (A+B+C) )
A = rocket launcher k2 adjusted AA value (48)
B = rocket launcher k2 count group (group 1 = 70, group 2 = 110, group 3 = 150)
C = ship class bonus (Ise-class = 70, other = 0)
```
I would like to display this proc chance, but ran out of space, so I used existing aaEffect
image

### DIFF
--- a/src/library/objects/Ship.js
+++ b/src/library/objects/Ship.js
@@ -2409,6 +2409,19 @@ KC3改 Ship Object
 	};
 
 	/**
+	 * To calculate Rosa K2 trigger chance (for now)
+	 * We need adjusted AA of ship, number of Rosa K2, ctype and luck stat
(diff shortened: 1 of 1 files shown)


sorewachigauyo:
Er, would it be better to `30 + 40*rosaCount` instead of `groupValue[rosaCount-1]`?

sinsinpub:
who know... only tanaka and formula inventor? 😹

PR #2530, opened 2018-04-04
Added ship level filter to strategy room ship list

ptyang:
Strategy room's ship list now has the feature to filter by ship level on minimum and maximum limits.

1) Added "Start Level" and "End Level" input fields to the right of the existing "Show by Name" input field.
2) "Start Level" input field will dictate the minimum level that a ship must be in order to display in the list. (Level inclusive, so the list will include ships this level and higher.)
3) "End Level" input field will dictate the maximum level that a ship must be in order to display in the list.
(Level inclusive, so the list will include ships this level and lower.)
4) Both fields do not need to be specified - if both are specified then both filters will be applied, if one is specified then only one filter will be applied.
5) If the start level is greater than the end level, then an error will display via turning the level specified red. Entering non numeric value for a level will treat it as level 0 for this check.
6) The ship level filter was added on top of the existing ship name filter, but the function was renamed. This means the ship level filter should work with the other filters as well.

### DIFF
--- a/src/pages/strategy/tabs/ships/ships.css
+++ b/src/pages/strategy/tabs/ships/ships.css
@@ -100,6 +100,49 @@
 	color: #f00;
 }
 
+.tab_ships .ship_level_filter {
+    float: right;
+    height: 18px;
+    margin: 0px 0px 2px 0px;
+}
+.tab_ships .ship_level_filter .level_start_label {
+    float: left;
+    font-weight: bold;
+    font-size: 12px;
+    padding-right: 5px;
+    line-height: 18px;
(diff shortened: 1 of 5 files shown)


ptyang:
Thanks for the comments. I know it wasn't safe but that's the only visible location I could retrieve the ship level from as far as I could tell without digging further into the code.

As for the isInRange function, that's cool - I didn't realize there was an existing function to do that. 👍 

I wasn't aware that the settings were persisted. Hope to see a fleshed out version at some point! And thanks for allowing my small contribution. :)

PR #2549, opened 2018-04-22
fix quests bug

jacky2001114:
When unselect or submit quests without opening KC3Kai , open KC3Kai will pile up all the quests together as this [image](https://i.imgur.com/zpx1G95.png) , some submitted quests even cannot be disappear.

Therefore, resetting localstorage.quests when opening KC3Kai can solve this problem.


sorewachigauyo:
localStorage.quests contains all quests, including completed ones
If you set all quest status to 1, you are resetting completed quest progress as well
Dependencies like https://github.com/KC3Kai/KC3Kai/blob/master/src/pages/strategy/tabs/flowchart/flowchart.js will be affected
Might want to look at https://github.com/KC3Kai/KC3Kai/blob/master/src/library/managers/QuestManager.js for active quests

jacky2001114:
You mean only set the uncompleted quests to status to 1?

sorewachigauyo:
That is one possible way of doing it, like what other viewers are doing. Though, having all your current quest reset to uncompleted every time you start the game feels inconvenient for the user.

By the way, user can click on 'box' beside quest to mark it as completed, to remove quests manually.

jacky2001114:
I am sorry that i don't know how to use the 'box' to remove the quests.
When starting the game , the tool panel will display the last-time quest list. The quest list changes when user click the quest box in game . Therefore, I think it is better to remove the extra quests manually and let the user get the correct quest list

sinsinpub:
write up some complex ideas here for whoever wanna implement auto quest completion enhancement:
(btw ideas are not mine, just collected from other places)

1. Set prerequisite quests to completion for same period quests (daily, weekly, monthly)
   * When handling `api_get_member/questlist` array, if a quest is period quest, and it appears (no matter state is opened or completed), its direct prerequisite period quest could be safely set to completion. 'prerequisite quest' can be gotten via walking through our `quest.json` meta property `unlock`. And nested looping possible.
   * For example: when quest `Bd5` is found in quest list, if there is `Bd2` and (or) `Bd1` still activated in kc3 list, it (or they) can be set to completion. (only change status, without touching tracking counter)
   * But be caution for the 'same period' condition, for example, if `Bw2` found, can not assume that `Bd5` is completed, since it might be the day after `Bd5` exactly completed.
   * Can not apply this to one-time quests, because `unlock` relationships are not game API data, they are manually updated by us (source from wiki), can not be guaranteed their accuracy.
   * On the opposite, if `Bd1` is not completed, but `Bd5` is still in kc3 list, either possible to change the `Bd5`, but not set to completion, should remove its entry, which meaning it is not opened yet.
   * This method is useful to clean period quests which are done at other places, can be triggered on any quest list page. The more pages viewed, the more related quests might be cleaned.
2. Set quests not existed in currently all available quests to completion
   * The key point is how to get all available quests. since `api_get_member/questlist` is a paginated list, it's not possible to get all quests for current player, from the result of only 1 API call, until player has viewed all quest pages.
   * We might cache the quest IDs on player viewing pages of quest list. When the cached amount of quests reaches the `api_count` (indicating total available amount of quests for current player), we can check current kc3 list and set those IDs to completion which not included in our cached ID list.
   * Cached quest IDs should be cleared for these scenes at least:
      * Player go back to home port;
      * Current computer clock has passed the daily quest reset time;
      * Perhaps more?
   * This method can be applied to all types of quests, but shortcoming is: can be triggered only when all quest list pages are viewed.

PR #2556, opened 2018-04-30
 Add anti-installation (and imps) additive/multiplicative bonus calculation

sorewachigauyo:
Split enemies into currently known installation types and assign bonus based on type

1. normal enemy
2. Soft-skinned installation
3. Pillbox
4. Isolated Island Princess
5. Harbour Summer Princess Damaged
6. Supply Depot Princess
7. PT Imp (not included with anti-installation modifiers)

~~Add bonus for Northernmost Landing Princess~~ (Should be the same as normal ritou anyway)
- [x] Add bonus for Harbour Summer Princess 
- [x] Add two Ka-Mi tank interaction
- [x] Add Toku Daihatsu + 11th Tank Regiment interaction
- [x] Precap/Postcap attack power integration

Some of these bonus interactions (like average improvement levels) are speculative, also, lacking info on certain equipment like 11th Tank Regiment.

Will probably be used to sim damage for future abnormal damage checker

### DIFF
--- a/src/library/managers/GearManager.js
+++ b/src/library/managers/GearManager.js
@@ -40,6 +40,26 @@ Saves and loads list to and from localStorage
 		jetBomberSteelCostRatioPerSlot: 0.2,
 		// steel_consumption = floor(api_cost * current_slot * 0.2)
 
+		// Anti-installation land damage per type and improvement
+		// Array format is [t2Bonus,t89Bonus,normalBonus]
+		landingModifiers : {
+			// Pillbox
+			0: {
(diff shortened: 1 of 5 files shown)


sorewachigauyo:
if api_name is undefined, current build will cause error, forgot to save before pushing

sinsinpub:
PT imps only 4 ids now. although game actionscript also uses 'api_name.includes' method, but i prefer id list 🤔

sinsinpub:
finally it goes like this if so many effects appear:
image

PR #2558, opened 2018-05-04
Defaulted MS YaHei to the Chinese Default Font.

kololz:


c933103:
Why YaHei?

sinsinpub:
i guess not everyone like this change. not sure about current win 10/osx versions, but YaHei is not the default font installed for tcn language in the past.
https://r12a.github.io/scripts/fontlist/

mikenzb:
Should be noted, PR closed and moved to #2559

PR #2559, opened 2018-05-06
Add new language "Cantonese"

kololz:
Yeah probably all things are done...maybe...just maybe


kololz:
yes, kinda.

c933103:
還是問一下為什麼會選擇把雅黑排到正黑前面？

sinsinpub:
iirc @kololz mentioned: `look better when used with English` at discord.
in fact, to achieve this, just putting some 'good-look and common' English fonts before `JhengHei` will work. currently `Helvetica` will not work on Windows platform, might just move `Arial` next to it.

so i also suggest to go back `JhengHei`.

PR #2563, opened 2018-05-08
Update Translation.js

c933103:
Use ISO-639 macrolanguage and language tag according to BCP47 guideline


c933103:
https://bugs.chromium.org/p/chromium/issues/detail?id=841464 Not sure if I word the report correctly

c933103:
The chromium bug tracker now say: We were unable to find repro steps for this bug as per c#3. If you have more data to reproduce this bug on chrome latest versions(M70-70.0.3538.110 or M71) or have clear repro steps, please reopen or file a new issue.
Thanks!

PR #2566, opened 2018-05-13
Update anti-installation modifiers

sorewachigauyo:
New info mostly from https://twitter.com/T3_1206/status/994258061081505792

Add new installation type: Supply Depot Summer Princess
Switched WG42 additive calc to array based method
Switched dummyId for Supply Depot Princess to damaged form (for icon purposes)
Add known toku daihatsu bonus
Add and show chuuha damage modifier
image
Update precap modifier to take into account simulated damage status

Also, is shipObj updated only after equip popup? Noticed weird numbers when equipping guns and unequipping with installation gear


sinsinpub:
i guess u have noticed... yep there is a known issue:
game api calls is in such order: change equip -> update ship all info.
so when popup triggered, the equip slots id updated, but stats from equip are not yet. thats why the oasw part i pass the `aswDiff`, fire power change might be needed too.

PR #2600, opened 2018-07-19
Air raid defense shotdown prelim calculation

sorewachigauyo:
Added preliminary calculation for land base defense in air raid from https://kancolle.wikia.com/wiki/Land_Base_Aerial_Support#Enemy_Raid_2
Added mouseover tooltip to display slot shotdown numbers when land base is set to defense
image

AS is used rather than AS+ since most event situation is either AS or AP for perfect defense
"LandBaseDefenseShotdown" : "Shotdown under AS\nSlot 1: {0}\nSlot 2: {1}\nSlot 3: {2}\nSlot 4: {3}",

### DIFF
--- a/src/library/objects/LandBase.js
+++ b/src/library/objects/LandBase.js
@@ -122,5 +122,43 @@
 		}
 		return returnObj;
 	};
+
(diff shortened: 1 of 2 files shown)


sinsinpub:
@sorewachigauyo current amount: `planes.api_count`, slot max: `planes.api_max_count`

PR #2646, opened 2018-09-02
Add unexpected damage prediction based on damage simulation

sorewachigauyo:
Tweaked battle prediction module to add related info for damage simulation
Simulate damage per attack basis based on battle status (formation, engagement, current health, cut-in, etc) and compare to actual damage
Display abnormal damage on the panel

There are a few problems with damage simulation, same cut-in number can have different damage modifier, like FBA/BA CVCI, NBCVCI or new DD cut-ins and some modifiers like anti-installation are still being researched. For cut-in damage, it is possible to (somewhat) determine exact cut-in type by checking api_si_list

This will (hopefully) be later used for tsunDB submission. Also, I'm not too sure working with the battle prediction module, getting some errors while running npm test


sorewachigauyo:
Can prepare a virtual ship object through sortie table, but it does not have ammo or slots, slots not so important because only CVNCI really depends on it (and is usually capped), can load default from master api

Ammo status can be somewhat guessed by going through node by node, but some details unknown - like sub node consumption

PR #2683, opened 2018-10-02
Show ship lock count on top of locking tab

HitomaruKonpaku:


HitomaruKonpaku:
@sinsinpub Whoops 

Lemme remove that

PR #2694, opened 2018-10-20
Display equip bonus if new equip has bonus or is synergy gear

sorewachigauyo:
Shows up like
image

Need some help with equip check logic
Thanks @DynamicSTOP for helping with boxes and layout

Made a typo with multiple equip count, whoops


Tibo442:
For missing / changed asw/los/evasion, lv 99 can be extracted from
ship3/port api [1]. Some lv 1 can be extracted from their respective album
entry.

DynamicSTOP:
That screenshot is outdated. Stats box text should be grey `color: #ccc;` and bg colour slightly different  :thinking: 
Also i am not sure if it would be looking ok for 4 "synergy" icons. Need to check if any of those exists.

PR #2704, opened 2018-10-30
Update ship list sort with in-game ordering

HitomaruKonpaku:
image

### DIFF
--- a/src/pages/strategy/tabs/ships/ships.css
+++ b/src/pages/strategy/tabs/ships/ships.css
@@ -373,6 +373,19 @@
 	width:44px;
 	text-align:center;
 }
+.tab_ships .ship_header .ship_item .ship_sortno {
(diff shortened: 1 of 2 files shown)


sinsinpub:
merged into `update-cumulative`
`#` means 'sequence number', might confuse ppl. i moved it before ship type, abbr with `C`, and updated the sorters handling.

info for who not familiar:
during game phase 1, the image was ordering by ship type (`api_stype`) first, then other ids. it was already implemented by our ship list when sorting by `Type`.
entering phase 2, they have changed it to `api_sort_id` in the first time, then level desc, member id asc. `api_sort_id` is an index combining ship class (`api_ctype`) with old sorting index (`api_sortno`).

HitomaruKonpaku:
Oh thanks for the update 👌

PR #2707, opened 2018-11-06
* export ships to kc3.moe

DynamicSTOP:
moved it here instead of showcase page because
- showcase export button handlers looks too nice and i don't want to break those with ugly ifs
- item export isn't even started
- it's "alpha" and i don't want to mess with current "helping" processes too much, while ui still isn't polished

### DIFF
--- a/src/pages/strategy/tabs/showcase/showcase.css
+++ b/src/pages/strategy/tabs/showcase/showcase.css
@@ -45,6 +45,9 @@
 .tab_showcase .exportResults a{
 	cursor: pointer;
 }
+.tab_showcase hr{
+	margin: 10px 0;
+}
 /* SHIPS
 -----------------------------*/
 .tab_showcase .stype_box {
--- a/src/pages/strategy/tabs/showcase/showcase.html
+++ b/src/pages/strategy/tabs/showcase/showcase.html
@@ -80,6 +80,10 @@
 	<div class="control hover" id="eventShipList">European ships</div>
 	<div class="clear"></div>
 	<div class="exportResults"></div>
+	<div class="clear"></div>
+	<hr>
+	<div class="control hover" id="exportToKC3_moe">Export to KC3.moe</div>
+	<div class="clear"></div>
 </div>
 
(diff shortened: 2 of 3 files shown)


sinsinpub:
@DynamicSTOP 
the message event listener will be added to `window` instance of kc3 sroom tab, not the newly opened `window` of kc3.moe. therefore `2 clicks -> 2 new tabs` obviously not solve (even irrelevant to) the duplicated listeners issue.

DynamicSTOP:
In any case double click is blocked now. Should be one listener at max at a time.
Less objects is better, removed ports from both sides. Not sure why i haven't chose that way half a year ago.

DynamicSTOP:
@sinsinpub 
anything else or should i merge?

sinsinpub:
image

DynamicSTOP:
Regarding error. On first load there is no options in local storage so that error is expected . It's triggers by [console.error here](https://github.com/DynamicSTOP/KCExport/blob/master/src/store/modules/options.js#L127) and i think i'll just add length check there later to suppress it until real problems happen. 


Names and length will be fixed later. I've just added [vue-i18n](https://github.com/DynamicSTOP/KCExport/commit/b49409bdff9a86412a009366532a13c359d19ab7) yesterday and i would need to find nice way to load only required langs at a time. There should be similar way to do this through webpack. Then i'll would be able to test names properly. I think German version should have longest words. :D 

Design isn't final at all, i still looking for better ways to display ships and made them look compact and nice. Hopefully Hitomaru will help.

DynamicSTOP:
@sinsinpub 
are you sure about hp?
default

and exported sees it properly
default

like in ship table
default

DynamicSTOP:
and i should probably fix sorting.... again...

sinsinpub:
sure, hp[0] is not master ship's hp[0], its instance's current hp. u need to damage ur ayanami first kappa

DynamicSTOP:
Ahhh... it's current hp. yeah, i should fix that...

PR #2709, opened 2018-11-14
TsunDB gunfit tests

Tibo442:
Automated submission of samples for purpose of (gun)fit verification.

Works by having some page in strategy room where testers can see which tests they can do, and choose one to do. This will make anyone able to do help out with these tests since barrier of entry previously was way higher (setting up logbook, calculating yourself etc). Moving the setup & calculation part away might make it more appealing to some. 

Tab:
Shows whenever or not the user has the required ship/level/equipment and disable the ones that are impossible to test currently.
Normal screen
When the tester has the ship(s) set up, it'll color yellow/green depending on if correct morale has been reached. 
Incorrect morale
Ready

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -972,7 +972,7 @@
 	},
 	{
 		"section": "SettingsTsunDBSubmission",
-		"help" : "",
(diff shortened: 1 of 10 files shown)


sorewachigauyo:
Yeah, library might be better
About saving initial morale to sortie, just thought about exporting sortie from history to replayer-simulator with node data in the future

PR #2712, opened 2018-11-17
Update list of required ships for quests A91 and B126

reinforce:
Yuugumo Kai Ni and Makigumo Kai Ni required for quests A91 and B126
This patch also fix Taigei quests chain


reinforce:
Oh, now i see changes in cumulative-update branch.

Issue #2733, opened 2018-12-31
[Suggestion]can i know nelson touch has been triggered in a battle?


yeshasuicide:
want to know nelson has touched or not

PR #2735, opened 2019-01-01
Add indicator for one-time sortie attacks

sorewachigauyo:
Add a purple background to shipboxes involved in special attacks (Nagato Cutin/Nelson Touch)
Highlight node letters where special attack has activated
Node letter highlight is only added on battle results

EDIT: realized that the purple bg blocks out the red taiha bg and could be potentially dangerous for players in the below case

EDIT2: fixed

Day battle example
image
Night battle
image

Resolves https://github.com/KC3Kai/KC3Kai/issues/2733

### DIFF
--- a/src/library/objects/Node.js
+++ b/src/library/objects/Node.js
@@ -662,6 +662,13 @@ Used by SortieManager
 			this.unexpectedList.push(...this.unexpectedDamagePrediction(result.fleets.playerEscort,
 				1, battleData.api_formation[0], battleData.api_formation[2], isRealBattle)
 			);
+
+			// Check for any one-time special cutin
+			const checkSortieSpecialAttack = attacks => attacks.some(attack => (attack.cutin || attack.ncutin) >= 100);
+			const playerMain = result.fleets.playerMain;
+			if (checkSortieSpecialAttack(playerMain[0].attacks)) {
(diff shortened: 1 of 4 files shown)


src/library/objects/Node.js
```diff
@@ -662,6 +662,13 @@ Used by SortieManager
 			this.unexpectedList.push(...this.unexpectedDamagePrediction(result.fleets.playerEscort,
 				1, battleData.api_formation[0], battleData.api_formation[2], isRealBattle)
 			);
+
+			// Check for any one-time special cutin
+			const checkSortieSpecialAttack = attacks => attacks.some(attack => (attack.cutin || attack.ncutin) >= 100);
+			const playerMain = result.fleets.playerMain;
+			if (checkSortieSpecialAttack(playerMain[0].attacks)) {
+				this.sortieSpecialAttack = playerMain.map(ship => checkSortieSpecialAttack(ship.attacks));
```
sinsinpub: btw, this one cannot check out the combined attackers since only flagship get 100/101

sorewachigauyo: BP module should already assign 100/101 to attackers involved

sinsinpub: yea, my fault. my tests always leave 1 enemy to be killed by nagato 😂

PR #2744, opened 2019-01-10
Add new ship order type for leveling tab

HitomaruKonpaku:
image


HitomaruKonpaku:
@sinsinpub Ya i know that but the file contain both type of indent style :(

PR #2747, opened 2019-01-13
Display special attack proc chance in ship tooltip

sorewachigauyo:
Night battle rates established on both wikia/wikiwiki https://kancolle.wikia.com/wiki/Combat/Night_Battle#Night_Cut-In_Chance and https://wikiwiki.jp/kancolle/%E5%A4%9C%E6%88%A6#nightcutin1

Day battle rates uses psvita formulae and may be unreliable, with CVCI introduction and combined fleet https://kancolle.wikia.com/wiki/User_blog:Shadow27X/Artillery_Spotting_Rate_Formula

image
Need some of these functions for upcoming tsundb submitter to get new type factors

### DIFF
--- a/src/library/objects/Fleet.js
+++ b/src/library/objects/Fleet.js
@@ -1318,6 +1318,24 @@ Contains summary information about a fleet and its ships
 		return result.every(v => v === -1) ? false : result;
 	};
 
+	/**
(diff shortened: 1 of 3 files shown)


sorewachigauyo:
Sure, sounds good

PR #2749, opened 2019-01-18
Add taiha warning popup on continue/advance

sorewachigauyo:
If player advances with taiha ship in fleet(s), will alert him via panel warning on `/next`

Not sure if we need another after having annoying taiha alert and red glow in panel, but maybe direct warning is more visible

image

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -739,6 +739,12 @@
 				"name" : "SettingsNoAnimation",
 				"type" : "check",
 				"options" : {}
+			},
+			{
+				"id" : "alert_taiha_continue",
+				"name" : "SettingTaihaContinueWarning",
+				"type" : "check",
+				"options" : {}
 			}
 		]
 	},
--- a/src/library/managers/ConfigManager.js
+++ b/src/library/managers/ConfigManager.js
@@ -121,6 +121,7 @@ Retrieves when needed to apply on components
 				alert_taiha_homeport : false,
 				alert_taiha_damecon  : false,
 				alert_taiha_noanim   : false,
+				alert_taiha_continue : true,
 
 				api_translation   : true,
 				api_tracking      : true,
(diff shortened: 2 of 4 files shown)


martirsadota:
I would like the modal to be more informative thoug, e.g.:

```
<ship(s)> (is/are) critically damaged!
You may want to end the sortie now by refreshing the page.
```

sorewachigauyo:
Sure, just tossing up the idea at the moment

sinsinpub:
tbh, i don't think a multi-task player who does't concern at game itself, will notice the panel alert...
it's possible to put the alert modal overlay to game screen, to prevent player eventually clicking the next/formation button, as long as he is not a web developer (page tamper). 😂

martirsadota:
That's also a possibility, though I'd really want that to be more pre-emptive. Right now though the only event that can be relied on is `next`, which is *after* proceeding.

Also, I guess an option can just be added to the **Annoying Taiha Alert** section for modal alerts on the screen, on the panel, neither, or both.

Tibo442:
There is `kcs2/resources/se/217.mp3` which plays when selecting yasen/continue, though that'll only work when user has SE's not muted ingame. 
Poi recently added 'battle end' notification which detects when `kcs2/resources/se/217.mp3` plays between battle start and battleresult api. For this purpose it would be between battle result and next/port api.

martirsadota:
I'll wager that most people who muted the game would have either muted only the BGM or are muting via Chrome/KC3 instead  of the in-game options, so that might work.

So it can be hooked to that event instead, but a reminder for the user to turn on SEs ingame should be displayed prominently (say on its settings item) to the effect of

```
IMPORTANT: for this option to work, please turn on SEs (sound effects) in-game. (How?)
```

then hovering on `How?` would reveal some instructions

```
To enable SEs, click the small gear icon on the lower-right corner of the home port screen, and drag the SE slider to the right.

If you want to mute the game, you may do so via the KC3 panel.
```

sorewachigauyo:
Taken over by https://github.com/KC3Kai/KC3Kai/pull/2819

PR #2760, opened 2019-02-05
Alternative Panel Theme

nonsensed:
While I'm still making small changes, I'm putting this up for testing or review in case there are problems with the implementation.

This adds higher resolution icons and a new panel theme, Moonlight. Moonlight features a slightly larger panel than Natsuiro, built specifically around supporting the new icon system and making some panel features provide a bit more information at a glance. It's designed more for people using higher resolution displays, but still works to varying degrees for lower resolution users, depending on how they have their panel set up. Red-green colorblindness was strongly considered for Moonlight's color scheme.

These changes also add more settings for customization in both Natsuiro and Moonlight. They were specifically made with extremely bright backgrounds in mind, for end-users with very little knowledge of css to be able to use any kind of image they want with as little extra work as possible. However, as I'm inexperienced, I feel there must be a better way to implement these options that I don't know about.

Link to an imgur album in order to avoid image spam:
https://imgur.com/a/pLuzH1f

The new icons are unoptimized. For quality reasons, I will attempt to keep them as close to their original specs as possible, but there are a few ways to reduce their size (and quality) as required.


nonsensed:
I want to avoid adding anything else for now, at least until other people have had a proper look at what ~~a terrible mess I've made~~ I've thrown together.

One thing to evaluate is adding stats to the LBAS tab in Moonlight. There's plenty of room to add additional stats, but I'm unqualified to pick what to do here exactly. Things like individual plane range, interceptor power, bombing power etc could be added depending on the plane type(?). I don't know what to do here. If someone wants to take a crack at this, please feel free.

Another thing is customizing the <100% zoom to work better, but this is something that can be worked on at a later time.

nonsensed:
I'm reevaluating a number of things based on the feedback I've received, so there will be further changes at some point soon.

nonsensed:
To-do: Address the problem of Asian font sizes being too small when playing at a low resolution.

nonsensed:
This is ready for testing or review I guess.
[fork's repo](https://github.com/nonsensed/tempotato)
[translation repo](https://github.com/nonsensed/kc3-translations)

I especially need feedback on the wide layout because I have no real experience with it or the types of setups people would use it for. It has been changed from being taller to being wider instead. It's basically impossible to retain the same height/width as Natsuiro without removing features that the tall layout has. The default background `wrapper.png` has not been adjusted to fit the layout yet. If we keep the wide layout's current dimensions, that will have to be adjusted.

The partner/waifu icon system was a fun thought experiment, but I lack the skill required to implement it. Will have to disable that completely before merging if nothing further is done with it.
[here is a video of that and most of the features that were added since the first pass](https://youtu.be/apIkhYXJ4v0)

nvm

PR #2777, opened 2019-03-07
Alternative Panel Theme Continuation

mikenzb:
### **Introduction**

This Pull Request attempts to accomplish three things:
1. It adds a new panel theme, _**Moonlight**_.
2. It implements a _**new ship icon system**_.
3. It adds _**several new options**_ for customizing either Natsuiro or Moonlight.

The Natsuiro theme should be minimally affected by all of these except for the new options, which enhance its ease of customization.

### **1. The Moonlight theme**

[Default panel](https://i.imgur.com/uhyHrex.png)
[Wide panel](https://i.imgur.com/cVQ1pZA.png)

Moonlight functions in fundamentally the same way as Natsuiro does, but it also differs from Natsuiro in a number of ways. Its primary purpose is to support the new icon system.

However, there is a degree of _**customization**_ built-in so that players can change displays to their liking. Currently the implementation is very minimal, but if there are no problems, there is a lot of room for expansion of features. Examples: [1](https://i.imgur.com/9SGgQBb.png) [2](https://i.imgur.com/hymJqTA.png) [3](https://i.imgur.com/ykx6bQ8.png)

The [combined fleet view](https://i.imgur.com/Zvc2Joi.png) is revised. It now displays more stats.

The [LBAS section](https://i.imgur.com/7ob0tRN.png) is revamped. Plane names are displayed by default, with enough space provided for very long Western names. While there are currently almost no individual LBAS plane stats displayed, this can be expanded upon.

The **_default view uses slightly more vertical space_**. This is specifically geared toward players who use high resolution displays, but support is not necessarily lost for lower-resolution users, depending on their individual setups and expectations. The panel *will* be taller.

The **_wide view uses more horizontal space_**. This was necessary in order to not sacrifice any of the features from the default view. Wide view has issues as of right now for a large number of users who expect it to fit in the same way as it does in Natsuiro. Solutions for this are being evaluated.

The color scheme and overall layout is still very much subject to change based on feedback, but it should be considered ready for release if there are no problems.

### **2. The new ship icon system**

The new icons do not currently utilize any form of file size reduction, favoring quality for 1080p and 4k users.

They use a 3:2 rectangular format instead of the previous circle format. However, they are intended to be fully compatible with the circle format. The main methods used to accomplish this are `object-fit: cover;` and `border-radius: 50%;`

The new icon system does not have a default background, so it's necessary to add a background color if desired. There is a new setting for this available for end-users.

### **3. New or changed customization options**

The **_Panel Box Background Color was changed_**. Both Natsuiro and Moonlight have different default panel colors, so a checkbox was added to allow users to switch between a custom color or the current panel's default.

Several new options were added with **_end-user ease of use_** in mind. They are designed so that an absolute minimum of css knowledge is required. It should now be possible to make panels look much better with custom background images regardless of their inherent color balance, even if `Interface Opacity` is set to 0%.

A [Misc. Panel Background Color](https://i.imgur.com/H0M5ouL.png) option was added. This sets a background color into specific elements. The option is similar to customizing the Panel Box Background Color, except for other sections of the panel. Which elements are affected depends on the theme.

A [Text Shadow Color](https://i.imgur.com/8gLgvGg.png) option was added. This affects most of the text across the entire panel, providing a single option that can dramatically influence the overall color balance and legibility of displayed text.

A [Ship Icon Background](https://i.imgur.com/MAUXLPh.png) option was added. The new ship icon system doesn't have a default background. This option adds two things: a) a background color to the icon matching the color selected and b) a line around the edges of the icon.

An option to **_increase select font sizes in the Moonlight theme_** was added. This is geared for low resolution users.

[link to the entire imgur album](https://imgur.com/a/sLEndzw)
Previous PR: (#2760)


nonsensed:
Current status: waiting on review.

Before merging, the partner system will have to be fully disabled.

I'm going through every ship icon and revising them to fit the circle format properly. If this PR passes review, I will rush to finish the remaining icons. Otherwise, I'll finish them at my own pace. Current progress with this is around 50%.

nonsensed:
I've split the Moonlight theme's customization options from Natsuiro's. People who are testing this PR should take note that if you're using custom css or any of the panel options in Moonlight, you will need to migrate those settings over to the new Moonlight section in the KC3 options.

nonsensed:
I'm sure something will come up after I post this, but -

Unless something comes up, this fork is now at a good stopping point. I don't plan to make any notable changes until at least when the event starts. Testers seem to be mostly satisfied with the current layout.

nonsensed:
@sinsinpub I'm afraid to even ask... but could we merge this soon? Is there anything you want to do or can we go ahead? I have had various people testing this branch for a while, and no major issues exist that I'm aware of. People keep asking me when we will release this.

I was finally able to do the last test I wanted to make, and it passed the test.

PR #2798, opened 2019-04-18
Allow to ignore unlock ships of taiha alert

HitomaruKonpaku:


HitomaruKonpaku:
> config name might be a little long 😂

I just follow the pattern tbh 😂

PR #2800, opened 2019-04-21
Add DE modding submissions for TsunDB

Tibowl:
Uses KC3Network listener because Kcsapi.js will have handled the modernization by the time the (async) DB submittor gets it (can't get lvl/master id of consumed ships / mod difference).

### DIFF
--- a/src/library/modules/TsunDBSubmission.js
+++ b/src/library/modules/TsunDBSubmission.js
@@ -947,7 +947,7 @@
 			const ship = KC3ShipManager.get(data.rosterId);
 			const modFod = data.consumedMasterIds.map((id) => KC3Master.ship(id));
 
-			// Checks in RemodelUtils.calcPowerUpParams
+			// Checks in main.js#RemodelUtils.calcPowerUpParams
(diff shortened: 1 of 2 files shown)


src/library/modules/TsunDBSubmission.js
```diff
@@ -917,7 +942,38 @@
 			//console.debug(this.development);
 			this.sendData(this.development, 'development');
 		},
-		
+
+		processModernize: function(data) {
+			const ship = KC3ShipManager.get(data.rosterId);
+			const modFod = data.consumedMasterIds.map((id) => KC3Master.ship(id));
+
+			// Checks in RemodelUtils.calcPowerUpParams
```
sinsinpub: should remark this: `main.js#RemodelUtil.calcPowerUpParams`

Tibowl:
> perhaps should not be called as `lolimodfod` any more?

The api endpoint is called like that since it was planned beforehand; the table and endpoint should work now. `specialmods` could also work, but I don't know how quickly TsunDB can be changed; like see #tsundb-api in kc3 discord.

PR #2812, opened 2019-05-10
Display ship bonuses with equipment in ship library

sorewachigauyo:
Filter equipment bonuses that ship is valid for and display it in mstship. For simplicity, only single bonus from gear is displayed.

May need to add new terms

For white theme
image

image

For dark theme
image

Thanks to @DynamicSTOP


DynamicSTOP:
nb: i wanted to style it similar to tree when it was already done... may be in next version.
image
root elements would be equip and leafs\children would be synergy items (flags)

rip `position:absolute` totally screwed up accordion animation 😄 guess because those plus and asterisks elements were treated as separate layer

PR #2819, opened 2019-05-19
Tiny heroes

DynamicSTOP:
hero

3 modes:
- Disabled (default, same as now)
- SE detection (Pretty one with fairy)
- Battle result detection (plain text starting showing from result screen at bottom)

Heavily based on prediction system.

Tibo suggested to use PIXI injection which can replace both of those modes and wouldn't require SE to be enabled. Sadly, all i found regarding that was kcs2_injectable.js and i am not sure why it's disabled so i haven't added that for now. 🤔 

TODOs:
- translations
- damecons check
- escort flagship taiha check (Not sure if needs any action. It was never confirmed to be 100% safe)
- retreats check (I doubt i would be able to test those outside of event)
- maps like 1-6
- hide while making screenshot (screenshot some times lags for me and could take 10s for a new tab, so i think it's more important to keep it blocked rather hope for fast screenshot)


DynamicSTOP:
mkay, since it's more like WIP thing i switched it to develop.

DynamicSTOP:
hmm.. yea i see now. damecons can't be ignored. in current version if you have one on FS, whole check will fail.

DynamicSTOP:
- Regarding locking whole screen... but.. but can we still show fairy there?
- i've seen people shrinking game window to game tab size on streams. Button is possible of course but it will be over game frame is that case. Details regarding ships in taiha status can already be seen in devtools panel, right?
- There were situations where player don't select formations, like fleet of 3 or less ships. Also new orange "Ambush" nodes from winter 2019 event doesn't require formations at all. May be leave as is at least for unknown maps. 🤔

PR #2833, opened 2019-06-11
Update OASW condition

CirnoV:
Fletcher can OASW without condition. 😄


CirnoV:
aaci1
aaci2

@sinsinpub I think it's same as Johnston's AACI pattern

PR #2835, opened 2019-06-12
Add voices and player fleets to TsunDB friendlyfleet

gakada:
CRC should be temporary, right now it is required that `uniquekey` represent `fleet` (somewhat) uniquely, so CRC should be a bit better than just a sum.


gakada:
Yeah, added `api_voice_p_no` as well.

@Tsubakura if you want to merge, I think maybe it can be enabled for the last week of the event as an experiment, since by that time current friend fleet data should be finalized. Then next event can have reworked table, also based on results of this experiment.

PR #2838, opened 2019-06-18
Moonlight theme

nonsensed:
To be filled in at some point. Most people here already know what it is.

https://github.com/KC3Kai/KC3Kai/pull/2777


mikenzb:
Unless anyone has any concerns, I will be releasing Moonlight in about a month (August 15th-17th)

PR #2840, opened 2019-06-21
Natsuiro (troll): Display "Jervaited" image instead of regular drop icon

martirsadota:
When `KC3 Troll Comments` is on, the "no drop" icon in the Natsuiro panel becomes the Jervaited emote:

Jervaited


martirsadota:
I could probably make it appear in Maps and Events tabs, but I prefer keeping Drop Record as-is.

PR #2873, opened 2019-08-26
 Adding average level display on ship list page

Jebzou:
Just a small change adding a display of the average level of the ships listed in the Ship list tab.

image


Jebzou:
Ya, there's some issues with the multi sorter.
I really made this just for showcase, it doesn't really have something to do with game strategy but can still be nice to see how's your progression in the game.

Given that it's just information maybe I should move it to the bottom of the list like this ? 

image

sinsinpub:
if the values from filtered ships are not important, my suggestion is: put them to showcase page and leveling page
if ships filtering is important, add a new section, maybe between filters and multi-sorters or any other proper place, and make it fold/unfold-able, might be folded by default if there will be many lines (without put exp things into tooltip, more stats can be summed/averaged there)

Jebzou:
Something like this ? 

image

image

Jebzou:
Need the filtering for things like excluding non locked ships, group by certain class, ...

sinsinpub:
btw, would you like to be listed in fork contributor section of kc3 about page?
image
if yes, u may leave some links (like homepage, twitter, etc) to be showed there

Jebzou:
Sure why not. 
Here's my twitter page : https://twitter.com/Jebzou

PR #2920, opened 2019-12-01
Fix ledger not working with very large history

Binaryhalibut:
The kc3 ledger suddenly stops working if your resource history is extensive enough. This is because current behavior of using apply with newTotalBuffer calls unshift with as many args as there are items in your resource history. The javascript engine does not like this when that number starts reaching hundreds of thousands.

To fix, we slice newTotalBuffer and prepend it in batches.

### DIFF
--- a/src/pages/strategy/tabs/overlodger/overlodger.js
+++ b/src/pages/strategy/tabs/overlodger/overlodger.js
@@ -1,4 +1,4 @@
-(function(){
+﻿(function(){
 	"use strict";
 	/*jshint: validthis true*/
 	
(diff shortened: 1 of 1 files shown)


Binaryhalibut:
It's still laggy as hell, but at least it loads now for the 1-day/1-week ledgers instead of showing all zeroes. Didn't try longer ranges, year just hangs (but maybe it works and I just didn't wait long enough).

Binaryhalibut:
year/total is call stack overflowing somewhere else.

It's late, hold off on the merge -- I'll go through and fix the rest of the call stack problems tomorrow.

Binaryhalibut:
OK, this fully fixes ledger for very large histories. All ledger features are now working.

image

That's a navaloverall table with 255828 rows, by the way.

Loading the larger timespans ends up taking upwards of half an hour to update due to sheer size, and sometimes there's no clear indication that it's still working besides CPU usage and debug log. This is a separate issue anyways.

PR #2925, opened 2019-12-04
Tiny heroes

DynamicSTOP:
ref #2819 
Since @mikenzb merged translations today i think it's worth of opening new PR for this branch again.

TODOs:
- [x] Touch screen mode. 
- [x] Recheck work with Damecons
- [x] Check retreat mechanic and SE mode
- [x] Backward compatibilities. see https://github.com/KC3Kai/KC3Kai/pull/3053#pullrequestreview-535958373
- [x] Port to DMM frame mode and API link mode
- ~~Check and not block safe advance nodes? (eg to 1-6-Finish node)~~
- [x] Check and block after FCF mechanism confirm screen?

Regarding "Touch Screens". Since some people used to shrink browser height\width it might be easily missed in game tab (i generally don't like adding such things there), i think it would be better to add new button or popup inside devtools instead.


sinsinpub (approved):
Only tested on my side with fake Taiha flag, not on real Taiha scene. It will be good if someone test it under the 2 legacy game mode.

PR #2930, opened 2019-12-13
Update Strategy Room

HitomaruKonpaku:
- Reposition buttons in dev panel https://github.com/KC3Kai/KC3Kai/pull/2930#issuecomment-572582445
- Fix ship list statistic not update when using name, level filter
- Avoid error when loading ship in case ShipMaster is outdate


nonsensed:
.h control buttons
.v control buttons

I think this is fine, but some may prefer the `btn_resize` button in the `.h` layout. Changing the icon for `btn_alert_toggle` is a very good change to avoid confusion.

I'm not knowledgeable enough about the strategy room changes to make any comment.

PR #2937, opened 2019-12-24
Natsuiro: Change CF icon according to CF type

martirsadota:
The combined fleet icon on the fleet selection button and status area will change depending on the current combined fleet type.

Also added icons for non-combined fleet scenarios.

image


martirsadota:
updated lang, did that work tho 🤔

PR #2942, opened 2020-01-08
moonlight syncup with natsurio

nonsensed:
Some sync-up. Adds and enables the `taiha alert toggle` and combined fleet icon display change features from recent commits.

img

Question
@sinsinpub  @HitomaruKonpaku 
I've changed the js for the taiha alert toggle a little bit from `natsuiro`'s version. Is this ok?
```
		//Button to pause the Taiha Alert sound
		$(".module.controls .btn_alert_toggle").on("click", function () {
			if (critSound.paused) {
				$(".module.controls .btn_alert_toggle img")
					.attr("src", "../../../../assets/img/ui/alertfairy.png");
				NatsuiroListeners.Fleet();
			} else if (ConfigManager.alert_taiha_sound) {
				critSound.pause();
				$(".module.controls .btn_alert_toggle img")
					.attr("src", "../../../../assets/img/ui/alertfairytea.png");
			}
		});
```
Reasons for the change: 
- If pressing the `alert toggle` button, it will start the heartbeat sound no matter what the user's options are and no matter if there is a taiha or not. Running a `NatsuiroListeners.Fleet();` is a lazy way to tell the panel to go check first. It passes through the normal taiha check process this way.
- Added a check for if the taiha sound option is on in user settings
- Fairy icon change because orz 

I'm not sure if it's acceptable to do `NatsuiroListeners.Fleet();` here. Is there a better method or is this fine?

I have no confidence in anything that I did!


sinsinpub:
i guess yours are better. i never test that new toggle button locally since i never enable taiha sound 😺  i think they should be copied to natsuiro too

invoking `Fleet` event to recheck and play taiha sound again should be ok, as long as there is no apperceptive performance issue

nonsensed:
Thank you! I'll keep it as `NatsuiroListeners.Fleet();` for now unless there end up being performance problems for users, then.

Adding it to natsuiro is definitely something I was wondering about.

nonsensed:
@sinsinpub 
Sorry for my ugly merging. Feel free to make any changes if you wish.

PR #3037, opened 2020-09-27
[moonlight] new consumables display

nonsensed:
The consumable page system in `.module.activity` was removed from `moonlight`. Consumable display was relocated to an area where there is more room and then restyled.

Users can toggle between several different categories of consumables at their discretion by clicking buttons on the left side.

General layout:
consumable display

Compact layout:
compact layout


sinsinpub (commented):
i didn't see `useitem_blank.png` used somewhere...? post some screenshots to demo?

nonsensed:
> 
> 
> i didn't see `useitem_blank.png` used somewhere...? post some screenshots to demo?

You're right. It was left in for possible future use, but it can certainly be removed if desired.

sinsinpub (approved):
it's okay to keep it. tbh i wanted it earlier. it can be used for those IDs no icon assigned by game, instead of currently used `empty.png`. could u either make another 75x75 version?

PR #3073, opened 2021-01-04
Keep quest list scroll

HitomaruKonpaku:


sinsinpub (approved):
it's okay for me.
but question is out if just remove `.scrollTop()` method: why did u decide to scroll top in the past?
i guess better impl might be keeping scrollbar positions by every in-game left-side tabs, so that on tab switching, weird scroll position for different list length will not get in.

HitomaruKonpaku:
Hi, back then I think it would be better to force it on top, but actually no

PR #3089, opened 2021-02-06
Update ImgBuilder

HitomaruKonpaku:
Allow to export directly from sortie/event page
Add LBAS info export data


sinsinpub (approved):
1. `tcn` removed from support langs, may map it to scn or jp, instead of default en
2. change merge target to `develop` branch

HitomaruKonpaku:
@sinsinpub Done

PR #3113, opened 2021-04-24
Add event statistics to strategy room

sorewachigauyo:
Display user event data from sortie history and localstorage maps

image


sorewachigauyo:
image
Update for new stats and text format
I think that this is good to go for planned features

Bugs known: In battle DB user may get undefined HP value, unable to reproduce so far

sinsinpub (approved):

PR #3193, opened 2022-03-08
Add TsunDB event battle API submitter

sorewachigauyo:
This is an attempt to create a submitter for battle data with as little reliance on KC3 custom APIs such as the ship damage simulator to make it compatible with other viewers. We submit the raw battle API along with as minimal information as possible to reconstruct the battle.

Planned schema
```json
{
	"map": {
		"desc": "Map name, e.g. 53-5",
		"type": "str"
	},
	"node": {
		"desc": "Node edge number",
		"type": "int"
	},
	"difficulty": {
		"desc": "Chosen map difficulty",
		"type": "int"
	},
	"debuffed": {
		"desc": "Checks if the node is debuffed, see https://github.com/KC3Kai/KC3Kai/blob/master/src/library/objects/Node.js#L665 for implementation",
		"type": "bool"
	},
	"cleared": {
		"desc": "Checks if the map is cleared",
		"type": "bool"
	},
	"fleet": {
		"desc": "Contains information about the sortied fleet(s), LBAS and support",
		"type": "JSON",
		"fields": {
			"fleet1": {
				"desc": "Array of JSON containing ship information",
				"type": "JSON[]",
				"value": [{
					"desc": "Contains information about the ship",
					"type": "JSON",
					"fields": {
						"id": "Ship mst id",
						"lvl": "Ship level",
						"morale": "Ship morale pre-battle",
						"stats": {
							"fp": "Ship visible firepower",
							"tp": "Ship visible torpedo power",
							"aa": "Ship visible anti-air",
							"ar": "Ship visible armor",
							"ev": "Ship visible evasion",
							"as": "Ship visible anti-submarine stat",
							"ls": "Ship visible line-of-sight stat",
							"lk": "Ship luck"
						},
						"equips": "Array of mst ids of ship equipment, -1 if slot is empty, extra slot is included",
						"improvements": "Array of improvement level of ship equipment, -1 if slot is empty, extra slot is included",
						"proficiency": "Array of proficiency level of ship equipment, -1 if slot is empty, extra slot is included",
						"slots": "Array of plane slot count, corresponds to api_slotnum"
						"flee": "Boolean to check if ship has been retreated by FCF",
						"fuel": "Fuel of the ship pre-battle",
						"ammo": "Ammo of the ship pre-battle"
					}
				}]
			},
			"fleet2": "Follows fleet1 or null if not in combined fleet",
			"support": "Node/boss support fleet, null if no such fleet assigned",
			"lbas": {
				"desc": "Array of JSON containing land base information",
				"type": "JSON[]",
				"value": [{
					"desc": "Contains information about the land base",
					"type": "JSON",
					"fields": {
						"planes": "Array of mst id of each plane in the base",
						"slots": "Array of plane count of each plane in the base",
						"proficiency": "Array of plane proficiencies for each plane in the base",
						"improvements": "Array of plane improvements for each plane in the base, -1 if none",
						"morale": "Array of plane morale for each plane in the base pre-sortie",
						"strikepoints": "Array of int[2] of edges targeted by the land base"
					}
				}]
			},
			"fleettype": {
				"desc": "Combined fleet type, corresponds to api_combined_flag",
				"type": "int"
			}
		},
	},
	"resupplyused": {
		"desc": "Checks if resupply has been used in the battle, obtained from request.params.api_supply_flag",
		"type": "bool"
	},
	"playerformation": {
		"desc": "Formation ID chosen by the player, obtained from request.params.api_formation",
		"type": "int"
	},
	"rawapi": {
		"desc": "Raw battle data",
		"type": "JSON"
	},
	"amountofnodes": {
		"desc": "Amount of edges in the map",
		"type": "int"
	},
	"apiname": {
		"desc": "API endpoint for the battle, e.g. api_req_sortie/battle",
		"type": "str"
	}
}
```

This should deprecate `TsunDBSubmitter.processUnexpected`. It looks like `TsunDBSubmitter.processEventAccuracy` uses 6-5 boss node and vanguard formations outside of events so this new method does not supersede that.

@Divinity123 @fourinone41 Could you check if this implementation matches your needs?

TODO:

- [x] Test LBAS cases (plane on cooldown, etc)
- [x] Test support fleet cases


Divinity123:
The schema looks good to me. I wonder if it is worthwhile to add a field for visible stats under "fleet" to make data processing easier (especially if we want to look at data from a long period of time and there is a change to visible bonus in-between).

Unless @fourinone41 has an objection, I am fine to stop collecting data from 6-5 and vanguard formation in normal worlds. We have enough samples for those and we are just waiting for data from this event to confirm our findings.

sinsinpub:
simple question: motivation of this is just for more samples from other viewers?

and do we have to wait for this pr merging on the very next release?

sorewachigauyo:
> simple question: motivation of this is just for more samples from other viewers?

Yes, I would like to extract more data from other viewers. Other than that, I want to reduce dependency on KC3 BP with increasingly complicated battle mechanics like submarine touch attacker order. I also want to reduce the need for new submitters/to change old submitters by extracting relevant information straight from raw battle API. 

> 
> and do we have to wait for this pr merging on the very next release?

No.

fourinone41:
> I wonder if it is worthwhile to add a field for visible stats under "fleet" to make data processing easier (especially if we want to look at data from a long period of time and there is a change to visible bonus in-between).

Yeah I agree there is value to submitting equipped stats (especially ev), otherwise we would have to track stat buffs/bonus changes manually. If concerned about being minimal, I think at least fp/tp/ev should be included.

Also naked stats (besides lk) may not even be needed under "fleet" because they're already in battle data's api_fParam.

> Unless @fourinone41 has an objection, I am fine to stop collecting data from 6-5 and vanguard formation in normal worlds. We have enough samples for those and we are just waiting for data from this event to confirm our findings.

IMO it's too early to disable, we don't even have a good model for vanguard evasion yet. More data is always useful, so unless there's an immediate need to disable I'd rather keep them. At least, we should actually do the confirmations first since unexpected things may always come up.
Fine to be old eventaccuracy format or this new format, as long as we submit the data.

sinsinpub (approved):

sinsinpub:
ready to merge?

sorewachigauyo:
Yeah, I will deprecate the other tsundb methods `processUnexpected` and `processEventAccuracy` after the next event when everything is confirmed to be working as intended.

PR #3231, opened 2022-07-05
Add additional backup/import method for OOM issues

sorewachigauyo:
With large backup sizes, users are getting OOM errors due to having to load the large backup file in RAM
To avoid this, I propose trying to stream the internal DB data in batches for export and import

**Pros**
- Reduces memory footprint
- Supports incremental backup, avoiding full backup in current process

**Cons**
- Exporting process for full backup is slower than current method
-  Incompatibility with existing backups, proposed backup file format is delimiter separated while current backup file format is a complete JSON

For exporting, we crawl through the DB in batches (batch size TBD). Each DB entry is then stringified and added to the backup file as a new line. We create a [WritableStream](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemWritableFileStream) for streaming the entries into the backup file.

For importing, we [read](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/read) each backup file in chunks. In each chunk, we separate the entries by the newline delimiter, parse into JSON and add into the DB tables.

TODO:

- [x] Add incremental backup procedure
- [x] Prettify and add details in HTML page for new backup functions
- [x] Determine a good number for backup batch size
- [x] Add error handling on DB import fail


sinsinpub (commented):
Since FS Access API supported since chrome 86, keep in mind to let old codes still runnable and handle the exception on lower version (ofc give disclaimer on page about that

sorewachigauyo:
image
WIP landing

PR #3238, opened 2022-07-27
Adding new maps in eventBattle submission and new fields in spAttack submission

Divinity123:
See the descriptions of the commit messages.

### DIFF
--- a/src/library/modules/TsunDBSubmission.js
+++ b/src/library/modules/TsunDBSubmission.js
@@ -1365,10 +1365,10 @@
 					// Uses raw api array instead to include engagement
 					misc.formation = (thisNode.battleDay !== undefined) ? thisNode.battleDay.api_formation: thisNode.battleNight.api_formation;
 					// Adding attacker HP for hit/miss prediction
-					template2['ship'].hp = attack.hp;
-					template2['ship'].maxhp = ship.hp[1];
+					template2.ship.hp = attack.hp;
+					template2.ship.maxhp = ship.hp[1];
 					// Adding position for calculations in vanguard formation
-					template2['ship'].position = [shipPos, fleet.ships.filter(id => id > 0).length];
+					template2.ship.position = [shipPos, fleet.ships.filter(id => id > 0).length];
 					misc.isCombined = isCombined;
 					misc.enemy = enemy;
 					misc.eposition = target;
--- a/src/library/modules/TsunDBSubmission.js
+++ b/src/library/modules/TsunDBSubmission.js
@@ -1364,9 +1364,10 @@
 					if (Object.keys(misc).length === 0) { continue; }
 					// Uses raw api array instead to include engagement
 					misc.formation = (thisNode.battleDay !== undefined) ? thisNode.battleDay.api_formation: thisNode.battleNight.api_formation;
-					// Adding attacker HP for hit/miss prediction
+					// Adding attacker HP and fleet type for hit/miss prediction
 					template2.ship.hp = attack.hp;
 					template2.ship.maxhp = ship.hp[1];
+					template2.ship.fleetType = this.data.fleetType;
 					// Adding position for calculations in vanguard formation
 					template2.ship.position = [shipPos, fleet.ships.filter(id => id > 0).length];
 					misc.isCombined = isCombined;
--- a/src/library/modules/TsunDBSubmission.js
+++ b/src/library/modules/TsunDBSubmission.js
@@ -1095,8 +1095,24 @@
 		processEventBattle: function(http){
 			// Sends the raw battle api with minimal info to reconstruct battle on server side
 			if(!this.currentMap[0] || !this.currentMap[1]) { return; }
-			if(!KC3Meta.isEventWorld(this.currentMap[0])) { return; }
 			const thisNode = KC3SortieManager.currentNode();
+
+			// Adding 6-4-C/J/K and 6-5/7-4 boss nodes for validation of the LBAS hit rate formula (https://twitter.com/Divinity_123/status/1538517358406418436)
+			const isNode64CJK 	= (this.currentMap[0] == 6 && this.currentMap[1] == 4) && ([3,10,11,18].includes(thisNode.id));
+			const isNode65M  	= (this.currentMap[0] == 6 && this.currentMap[1] == 5) && ([13,18].includes(thisNode.id));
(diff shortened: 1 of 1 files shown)


sorewachigauyo:
The `spattack` changes look okay, but I am a bit hesitant about extending `eventbattle` to normal world maps. The implementation looks okay and I don't believe there are any filters on the API side to prevent normal world submersion. Though, other than it going against the naming, at design, I aimed for the following for `eventbattle`:

1) Having a lighter memory footprint than `abnormaldamage` for the purpose of storing and analyzing battle data in event maps
2) Able to be easily implemented by different KC viewers, with minimal to no adjustment after inception for low upkeep
3) Ease of maintenance to dump data on an event-by-event basis if needed.

This local change does not mean that EO has to accommodate this submitter change. I understand that this change might allow you to collect more data for LB analysis, but I am not so sure about the futureproofing.

Divinity123:
I do agree with your points.
The intention fourinone and I have is to record samples with LBAS in normal maps for a reasonably short duration just to accumulate enough samples for us to validate our LBAS hit rate formula. Since we already have the code for `eventbattle` submission, we might as well use it.

If we do want to keep `eventbattle` for event maps only, I propose to create another table in TsunDB to store 'scratch data'. The idea is that we may want to collect data from specific nodes/maps in normal worlds for validation projects from time to time. And we can make submissions to this table for that purpose.
To keep the submitter simple, we can use the same code as that for `eventbattle` and change the filter at the beginning (maybe add a tag to indicate which validation project the submission is for).

In the long run, I do think a submitter similar to `eventbattle` can help validation projects greatly, as it makes data more accessible (as opposed to single attacks only in `eventaccuracy`). Projects like LBAS hit rate formula and CF targeting logic won't have been possible without the raw APIs we have access to.

Please let me know what you think.

sorewachigauyo:
The proposal sounds good to me with regards to futureproofing and storage duration.

We actually have an existing but unused table `gunfits` with validators for each [gunfit test](https://github.com/Tibowl/TsunTools/blob/master/config/gunfits.json) so there is precedence for such a system. Perhaps you can come up with a proposal of the validation system and we'll discuss more details from there.

Divinity123:
Sorry for the late reply, I had a few really busy days. I have reverted the changes in `processEventBattle`.

As for a new validation system, I am thinking about a system that is similar to `eventBattle`, where the whole rawapi is submitted to TsunDB and the validator(s) can then process the data locally. We only need to change the submission conditions when we need to run one such validation project.
In my opinion, it has the following benefits:
1. Easy to maintain on the backend.
2. Flexible for different validation projects, i.e. we don't have to write a new submitter every time.
3. Full data availability (we could not use the majority of NagaMutsu touch submissions because of the absence of HP information for example).
Ideally, we only need to turn on this submitter for one or two updates for each validation project. The data can then be downloaded by the validator(s) and deleted from TsunDB. We will need to figure out another way if the submitter takes a lot more space than expected on TsunDB.

sorewachigauyo:
Oh, by validation system, I meant validation for the submission conditions. Otherwise, that plan looks good to me. 

Should we proceed with this PR first then?

Divinity123:
To clarify, is your intention to use `eventbattle` for additional validation projects? I think creating another table for those is better because it is easier to maintain. I don't think we should use fine-grained submission conditions in this case as the memory footprint can be quite large.

I am happy to go with the PR for `spattack` for now.

sorewachigauyo (approved):

sinsinpub:
[84a334c] slightly adjusted 'location' for some properties, plz check if there is problem @Divinity123 
since i guess `fleetType` should not be belonged to ship context, so moved it to `misc` obj.
but actually there should be 2 fleet types, one for player, one for enemy (combined flag only).
anyway for now u can get combined flag from eposititon if targeted enemy is > 6?

sinsinpub:
another thing about amount of player fleet ships to clarify, for vanguard formation, when 1 or more ships sunk or retreated (FCF) during mid-sortie, will the ship amount and ship position be changed in the next battle?
afaik, FCF doesn't affect ship array length, nothing changed; sunk ship disappears from array in the next battle, should affect vanguard.

Divinity123:
> [[84a334c](https://github.com/KC3Kai/KC3Kai/commit/84a334c8e334e4fe7353f7f4899e773514b49082)] slightly adjusted 'location' for some properties, plz check if there is problem @Divinity123 since i guess `fleetType` should not be belonged to ship context, so moved it to `misc` obj. but actually there should be 2 fleet types, one for player, one for enemy (combined flag only). anyway for now u can get combined flag from eposititon if targeted enemy is > 6?

Thanks for the changes. I structured the properties in this specific way to make them compatible with [`eventaccuracy`](https://github.com/KC3Kai/KC3Kai/blob/84a334c8e334e4fe7353f7f4899e773514b49082/src/library/modules/TsunDBSubmission.js#L1449). More specifically, the properties should be structured as follows:
- Ship HP (`hp`)/ max HP (`maxhp`) are [separate entries](https://github.com/KC3Kai/KC3Kai/blob/84a334c8e334e4fe7353f7f4899e773514b49082/src/library/modules/TsunDBSubmission.js#L1405) (`eventaccuray` has ship HP as `attackerHP`, I used 'hp' in my own code)
- Ship position takes the shape [[`fPos, fCount, ePos`](https://github.com/KC3Kai/KC3Kai/blob/84a334c8e334e4fe7353f7f4899e773514b49082/src/library/modules/TsunDBSubmission.js#L1467)]. We do want `eCount` if the enemy is in vanguard formation, but we can get all the data from `eventbattle`.
- Fleet type belongs to [`ship`](https://github.com/KC3Kai/KC3Kai/blob/84a334c8e334e4fe7353f7f4899e773514b49082/src/library/modules/TsunDBSubmission.js#L1449)

We do not need an extra flag for enemy CF as `formation[1]` includes that information.


> another thing about amount of player fleet ships to clarify, for vanguard formation, when 1 or more ships sunk or retreated (FCF) during mid-sortie, will the ship amount and ship position be changed in the next battle? afaik, FCF doesn't affect ship array length, nothing changed; sunk ship disappears from array in the next battle, should affect vanguard.

I think you are right about this. This can be fixed if we use `api_f_nowhps.length` or something similar from the raw API to deduce `shipCount`.
The number of samples we have with sunk ships is small enough such that it has not affected our analysis yet, but it is always a good idea to use the right thing for futureproofing.

sinsinpub:
ok, will revert struct back to compatible with old

PR #3241, opened 2022-08-19
Add expedition only config for pushalert

Sean2525:
Hi, I've been using Pushbullet and Parsec for expeditions for a month.
But there was a problem that bothered me.
It also will send other notifications like construction, repair, etc.

I want the expedition notification only.
So I've added an option, as well as translations for EN, JP, and CN.
https://github.com/Sean2525/KC3Kai/commit/5a3b72ed5e6be1b1ac25dc3f290234f28c08f1c5
https://github.com/Sean2525/kc3-translations/commit/897087137869314bcb881d2f5e753c8e8d10ef9e

The notifId can be used to check if it is an expedition.
https://github.com/KC3Kai/KC3Kai/blob/408ca39eb817dcae852a8676069043e061b737ab/src/library/objects/Timer.js#L174-L179

image

After I tried it for a day, it works well.

Could you help review it? 
Thanks.

### DIFF
--- a/src/data/settings.json
+++ b/src/data/settings.json
@@ -1129,6 +1129,12 @@
 				"options" : {}
 			},
 			{
+				"id" : "PushAlerts_expedition_only",
+				"name" : "SettingsPushAlertsExpeditionOnly",
+				"type" : "check",
+				"options" : {}
+			},
+			{
 				"id" : "PushAlerts_key",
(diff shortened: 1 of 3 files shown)


sinsinpub (approved):
Looks good to me, please send ur translations PR too

PR #3299, opened 2023-09-23
Fixing CSS in ship locking planner

Nya-chan:
- **Removed limit on locked ship list height**
There was a dynamic height set on the table with ships, but the height got smaller over the years (with more and more locks introduced), leading to a super tiny window. This was probably to have ships closer to drag & drop area, but in my opinion this is not really necessary with solid filtering options available (and you can drag to top of the window still). Thus I removed the height limit for now.
- **Reworked locked ships area to support dynamic lock count in better way via CSS**
Removed some fixed settings and introduced automatic layout using flexbox features. Left the minimum width of one column to 70px (they in fact stretch to ~76px with the gaps in this layout when there is the maximum of 8 locks / columns). Put drag & drop features to bottom of each lock and increased the size of ship icons a bit, also ordering them automatically in the lock area using flexbox again (in some widths this looks better, in some the gaps are a bit large - when the lock is for example 2.9 icons wide, so the third one doesn't fit and 0.9 icon width goes into gaps -, but that's a price for it being automatic)

Example locked ships area from this event:
image


sinsinpub (changes requested):
removing 'height limit' is incomplete, 'Toggle scroll' button and its event handler still there... why not just swap the default? default to no limit  just do the same thing, and still possible to fixed height via toggle button.

plz change merging target branch to `develop`, as always.

sinsinpub:
I've rolled it back to use `margin` spacing instead of flex `gap`, working for old chrome versions, still fit in 2 rows up to 18 tags. do u accept this?
image

(height limit just clicking the toggle button to on/off, still be a thing)

Nya-chan:
> removing 'height limit' is incomplete, 'Toggle scroll' button and its event handler still there... why not just swap the default? default to no limit just do the same thing, and still possible to fixed height via toggle button.
Tbh didn't even (ever) realize such button was there, will switch the functionality (unless you already did that)

> plz change merging target branch to develop, as always.
Switched the target branch, mb

> sad news: gap was avoid intentionally since quite new chrome required.
Ch84 is three years old, are there really KC players or clients that use over three year old installs without restart / update? I don't think we sadly have the data for usage, I guess? But fair enough note.

> I've rolled it back to use margin spacing instead of flex gap, working for old chrome versions, still fit in 2 rows up to 18 tags. do u accept this?
Will check right away, thanks for the work!

Nya-chan:
@sinsinpub Looks good to me, tested around for a bit and couldn't find anything easy-ish to fix (could probably play more with the margins, but not worth it at the moment)

sinsinpub (approved):

sinsinpub:
😎 3 years look 'new' to me. 
kancolle game itself 10 years old, html5 5 years.
MSX0 comes out these days, just revived a computer 40 years ago.
Not everyone let things auto-updated (remotely controlled) by those IT companies.

PR #3381, opened 2025-11-16
New setting to exclude land bases from LBAS calculations

Myarnoid:
Adds a new setting to exclude land bases from the calculations for the LBAS panel, such as the indicator for resupply/morale as well as resupply/sortie costs.

Use case: 
Having a fighter sweep land base for farming world 7. This land base will be shaved down to 2/3 fighters and never resupplied. Currently (i.e without this new setting), the resupply/morale indicator for LBAS is permanently red, even though you'll never want to resupply that base. This makes it hard to see when you have to resupply your event land bases.

Proposed additions to terms: 
"SettingsLBASInfoCalculationsExcludeBases" : "Land bases excluded from LBAS info",
"SettingsLBASInfoCalculationsExcludeBasesTip" : "Set a list to exclude land bases from being considered for calculations in the LBAS panel, including the indicator for resupply/morale as well as sortie/resupply costs. Format: [base ID1, base ID2, ...]. For example: [62,71] means the second base in World 6 as well as the first base in World 7 will be excluded from the relevant calculations.",

### DIFF
--- a/src/library/managers/CalculatorManager.js
+++ b/src/library/managers/CalculatorManager.js
@@ -600,8 +600,10 @@
     const getLandBasesResupplyCost = () => {
         const total = {fuel: 0, ammo: 0, steel: 0, bauxite: 0};
         $.each(PlayerManager.bases, (i, base) => {
-            const cost = base.calcResupplyCost();
-            Object.keys(total).map(k => { total[k] += cost[k]; });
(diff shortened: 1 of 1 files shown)


sinsinpub (approved):

sinsinpub (approved):
will add terms to tl repo.
not a big deal, perhaps will shorten config key a bit, since it will take spaces of `localStorage`, probably `info_lbas_excl` enough

PR #3412, opened 2026-07-23
Performance optimizations for Strategy Room sortie history

HitomaruKonpaku:
## Summary

Strategy room performance overhaul: async sortie ledger loading, compound DB indexes, event map UX improvements, and shared settings UI across maps/event tabs.

### Performance

- **Compound indexes on sortie table** — Added `[hq+world]` and `[hq+world+mapnum]` compound indexes (DB v77). Refactored `count_world`, `get_world`, `count_map`, `get_map`, `count_normal_sorties`, and `get_normal_sorties` to use them. Extracted shared `get_sorties_with_battles` helper to deduplicate the sortie+battle join logic across all world/map/normal query paths. Normal sortie queries now use `world <= 20` instead of `world < 10`.

- **Async sortie ledger with Bottleneck** — `showSortieLedger` now returns a promise chain instead of a nested callback pyramid. Ledger queries are scheduled through a `Bottleneck` limiter (max concurrency 2) so the UI renders immediately while resource tooltips load in the background. Tooltip-heavy node rendering (`buildEnemyStatsMessage`, `buildSupportAttackMessage`, `buildAirPowerMessage`, `buildAirBattleLossMessage`) is also scheduled through a separate high-concurrency Bottleneck limiter. Added `leave()` hook to cancel all pending jobs on tab switch.

- **Queued/loading indicators** — `.sortie_map` cells get yellow `queued` and green `loading` border highlights during ledger fetch lifecycle.

### Event History UX

- **Settings bar** — Y/T/B/O toggle buttons moved from inline batch-toggles row into a proper checkbox bar above the map list, shared across both maps and event tabs. Old inline toggles commented out.

- **Ledger opt-in checkbox** — New "Ledger" setting (off by default) since ledger queries are expensive. Saves to `localStorage` under `srMaps`.

- **Event world selection persistence** — Last-selected event world+map saved to `srMaps` and restored on re-entry.

- **Boss HP sub-value display** — Transport gauge (type 3) `api_sub_value` delta shown inline next to boss HP, with hover tooltip showing the calculation.

- **Dropdown labels** — World dropdown options now show `[worldId] label` format for clarity.

- **Sortie end icon titles** — "Clear" and "Catbomb" icons get `title` attributes for discoverability.

### Refactor

- **`EVENT_WORLD_TERMS` → `KC3Constant`** — Moved from a local `const` in `sortielogs.js` to a shared `window.KC3Constant.EVENT_WORLD_TERMS`. Used by both event tab and eventstats tab.

- **`eventstats` world switcher** — Rewired to use `KC3Constant.EVENT_WORLD_TERMS` for consistent dropdown labels with translation support and `[id]` prefix, matching the event tab.

- **`Node.js` — Transport gauge data capture** — `api_landing_hp.api_sub_value` persisted to `eventmap` on sortie records so it's available for display in the history view.

- **Tab `leave()` lifecycle** — `strategy.js` calls `definition.leave()` on the previous tab before switching, giving tabs a cleanup hook.

### New dependency

- **Bottleneck** (MIT) v1.16.0 — Added `bottleneck.min.js` (~8 KB) for async concurrency limiting.

### Screenshot

image

image

image


src/library/modules/Database.js
```diff
@@ -581,91 +548,128 @@ Uses Dexie.js third-party plugin on the assets directory
...
-				.and(function(sortie){ return sortie.world == world && sortie.mapnum == map; })
+
+		count_world: function (world, filterFunc, callback) {
+			return this.con.sortie
+				.where("[hq+world]")
+				.equals([this.index, world])
 				.and(filterFunc)
-				.count(callback);
+				.count()
+				.then((value) => {
```
sinsinpub, line 621: this `.then()` meaningless, just keep `.count(callback)` is identical to invoke callback with count value and return the promise chain with count value as parameter too.

sinsinpub:
update PR desc and upload some screenshots to show how the changes look like, either including normal map history page

PR #3415, opened 2026-08-07
Performance and feature improvements for data import, SRoom (map/event ledger) data caching

HitomaruKonpaku:
**DataBackup — bulkAdd import (V1 & V2)**
- Dexie 1.5.1 compat (case-sensitive filename)
  - `dexie.min.js`
- `processDB_2`: chunked `bulkAdd`-based DB import replacing per-record `table.add()`
- `loadDataFromFolder_2`: same `bulkAdd` optimization for folder-based V2 import
- Merged `loadData`/`loadData_2` and button handlers via version param + shared selectors

**KC3Cache — in-memory LRU**
- Reduced cache overhead vs IndexedDB round-trips

**Strategy Room**
- Event clear mode filter dropdown


sinsinpub (changes requested):
Hmm, it's named `cache` but actually implementation by urself in a regular IndexedDB...without proper cache invalidation/expiration/clear mechanism, there would be kinds of unexpected results...
Only removing previous sortie's cached items should not be enough, easily imagine the cached values will be incorrect for such cases:
 * Don't resupply ships/LBAS at once after latest sortie, we know many ppl doing this, and once viewed ledger, later resupply consumptions would be never added up
 * Don't repair ships at once, similar with previous, many ppl keep ships to be repaired later
 * Sortie to normal maps between event maps, this is known to break LBAS cost too

And since it's just a regular idb, cumulative huge cache will still eat storage spaces and slow down everything someday...

I think it may not be worth to add extra complexity for us to upkeep such a 'cache' for only speedup consumption tooltips. As alternative way, added a click event to load single sortie's tooltip on demand, when batch load settings off.

Suggest to remove the cache part, leave this PR as dexie ver upgrade and backup updates?

HitomaruKonpaku:
I still think we should have a `cache`

Imagine user with `Resources` enabled + show 50 per page, switching around will be a huge pain

I believe this case exist (check around discord)

So instead of remove it, lemme implement the invalidation thing

sinsinpub:
Disabling `Resources` exists for that case, doesnt it? With disabled checkbox, switching around without pain, click the sortie to load resources concerned at, make more sense, less bugs.

HitomaruKonpaku:
> Disabling Resources exists

 I mean yes, but some prefer with it on so checking around will be easier 🤔 
 
 The click exist for ppl that don't check them as much, still a valid case

sinsinpub (approved):

